PR #17, opened 2019-01-24
Fixed overflow problem in fn clear_screen()

HugoRahlen:
Had some problems with my programme crashing due to stack overflow when clearing the screen (but not on mintty obviously). X and Y in csbi.dwSize are u16, but multiplying them easily overflow that since its looking at clearing the whole buffer and not just the window. The buffer is always 300 in Y in my cmd, and 1000 in my cmder.


HMate:
I have the same problem when running from powershell. Would appreciate if this was merged back. I tried out the fixed branch on my project and it did fix the error.

PR #20, opened 2019-03-17
Fix multiply with overflow at clear_screen

dalance:
If console window is large at Windows, `clear_screen()` becomes panic.
For example, my console is `csbi.dwSize.X` = 242 and `csbi.dwSize.Y` = 300.
So `cells` becomes 72600 and it overflows because the type of `cells` is `SHORT`(`i16`).


dalance:
Sorry. This is duplicated. #17

PR #57, opened 2020-04-27
Make it work (ish) on Windows 7

CreepySkeleton:


CreepySkeleton:
LOL

src/term.rs
```diff
@@ -403,21 +403,22 @@ impl Term {
...
 
-    #[cfg(all(windows, features = "windows-console-colors"))]
+    #[cfg(all(windows, feature = "windows-console-colors"))]
     fn write_through(&self, bytes: &[u8]) -> io::Result<()> {
         if self.is_msys_tty {
             self.write_through_common(bytes)
         } else {
             use winapi_util::console::Console;
 
+
```
pksunkara, line 413: Unneeded line.

src/windows_term.rs
```diff
@@ -382,7 +382,7 @@ pub fn set_title<T: Display>(title: T) {
     }
 }
 
-#[cfg(feature = "windows-console-colors")]
+#[cfg(all(windows, feature = "windows-console-colors"))]
```
pksunkara, line 385: You don't need this change since this file exists only in `windows`

CreepySkeleton: But it makes it explicit

CreepySkeleton:
Any idea about Invalid Handle?

pksunkara:
That's not related to this PR.

pksunkara:
Oops, looks like it is related to this PR, because of the `?` you added. The doctest isn't working because of that.

CreepySkeleton:
Well, if `Console::stdout/stderr` has returned an error, there's something very wrong with the environment already.

pksunkara:
Yeah, we can't rely on Console and will have to do the logic ourselves

PR #59, opened 2020-04-27
Reset order of size() tuple to height first

### LINKED ISSUE #58: Order of width/height in Term::stdout().size() has changed
The order of width/height in the tuple output of `Term::stdout().size()` has changed from 0.10.0 to 0.10.1. I assume that this was not intentional. Otherwise, 0.10.1 breaks semver.

Edit: The docs for `size_checked` explicitly say *"the terminal size in rows and columns"*, so I guess rows should indeed be first.

The order is still "wrong" in 0.11.

see: https://github.com/sharkdp/bat/issues/955

Long-term, using a simple `struct` instead of a tuple would be a good idea IMO.

sharkdp:
closes #58

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -31,7 +31,7 @@ pub fn is_a_color_terminal(out: &Term) -> bool {
 
 #[inline]
 pub fn terminal_size() -> Option<(u16, u16)> {
-    terminal_size::terminal_size().map(|x| ((x.0).0, (x.1).0))
+    terminal_size::terminal_size().map(|x| ((x.1).0, (x.0).0))
 }
 
 pub fn read_secure() -> io::Result<String> {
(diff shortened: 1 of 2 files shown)


pksunkara:
Release 0.11.1

Should I yank 0.11 and 0.10.1?

sharkdp:
Thank you for the quick release!

> Should I yank 0.11 and 0.10.1?

It seems like other people ran into this as well: https://github.com/dotboris/alt/pull/118#issuecomment-620252942

pksunkara:
Yanked. I will release a 0.10.2 with the fix though just in case.

pksunkara:
@sharkdp Was looking at your code and was there any reason you are using `ansi_term` to color instead of `console`? Would reduce one of your dependency. Also `ansi_term` doesn't work on windows console.

sharkdp:
It's mostly because we started with `ansi_term` very early on and only added `console` later when we needed the terminal width. We could try using `console` only, thanks!

PR #62, opened 2020-05-07
read only one byte in read_single_key unless we find an escape sequence

rjwalters:
When input is pasted into a unix terminal and collected using `read_single_key` I noticed that only every 20th character was available. I don't think it should be necessary to read more than one byte at a time. (Hopefully no unicode issues etc.)

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -89,7 +89,7 @@ pub fn read_single_key() -> io::Result<Key> {
         if read < 0 {
             Err(io::Error::last_os_error())
         } else if buf[0] == b'\x1b' {
-            // read more bytes if the first byte was the ESC code
+            // read 19 more bytes if the first byte was the ESC code
             let read = libc::read(fd, buf[1..].as_mut_ptr() as *mut libc::c_void, 19);
             if read < 0 {
                 Err(io::Error::last_os_error())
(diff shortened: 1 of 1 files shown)


rjwalters:
This causes problems for keystrokes that are longer than one byte... for example in dialoguer's Select functionality. I think we need to look for multibyte keystrokes and read more of the buffer when we find them...

``` rust
b"\x1b[D" => Key::ArrowLeft,
b"\x1b[C" => Key::ArrowRight,
b"\x1b[A" => Key::ArrowUp,
b"\x1b[B" => Key::ArrowDown,
b"\n" | b"\r" => Key::Enter,
b"\x1b" => Key::Escape,
b"\x7f" => Key::Backspace,
b"\x1b[H" => Key::Home,
b"\x1b[F" => Key::End,
b"\t" => Key::Tab,
b"\x1b[3~" => Key::Del,
```

pksunkara:
Sounds good, but how does utf8 work with this?

rjwalters:
> Sounds good, but how does utf8 work with this?

I'm not sure I understand the intent of `get_single_key()` -- it looks like it could be pretty complicated to properly support all of the [escape sequences](https://github.com/bminor/bash/blob/master/lib/sh/strtrans.c). Is it enough to make sure we manage the full set in `pub enum Key`?

We can detect UTF-8 byte sequences and manage these as another special case if that's in scope - does `termios` support UTF-8 input?

It would be nice to have some tests for this sort of thing if there is a way to mock input in termios.

rjwalters:
I added special cases to capture UTF-8 based on https://en.wikipedia.org/wiki/UTF-8

pksunkara:
Thanks. This is great work. Really appreciate it. Feel free to open PRs anytime you want. I will always respond fast. 😄

PR #68, opened 2020-05-29
Adding support for emojis on platforms other than macos

craftytrickster:
as long as they have unicode support. First proposed by github user arabidopsis


craftytrickster:
For issue https://github.com/mitsuhiko/console/issues/18

craftytrickster:
manually confirmed that it works on linux

PR #73, opened 2020-08-07
Remove termios dependency

AdminXVII:
Since the libc crate provides the same functions as the termios crate,
prefer to use it, as platform support is better (Redox, etc.)

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -50,7 +50,7 @@ pub fn read_secure() -> io::Result<String> {
 
     let mut termios = core::mem::MaybeUninit::uninit();
     if unsafe { libc::tcgetattr(fd, termios.as_mut_ptr()) } != 0 {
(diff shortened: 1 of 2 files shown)


src/unix_term.rs
```diff
@@ -48,10 +48,16 @@ pub fn read_secure() -> io::Result<String> {
         }
     };
 
-    let mut termios = termios::Termios::from_fd(fd)?;
+    let mut termios = core::mem::MaybeUninit::uninit();
+    if unsafe { libc::tcgetattr(fd, termios.as_mut_ptr()) } != 0 {
+        return Err(io::Error::last_os_error());
```
pksunkara, line 53: Can you please extract out this error returning into a function that returns `Result` and use the `?` operator more?

AdminXVII:
@pksunkara Done

PR #74, opened 2020-10-10
Return immediately if Esc is pressed

### LINKED ISSUE #66: Escape key is not recognized as correct Enum and not returned immediately
There are two (possibly related issues here)

When I press the escape key when running `term.read_key()`, nothing is returned until another (anything will do) key is pressed. 

Regardless of what the second key pressed is, the value returned is not `Key::Escape` but `Key::Char('\u{2b}')`

I was digging into the source code and noticed this message that could point to where the issue might lie?

https://github.com/mitsuhiko/console/blob/9aab753cc811b3fed824c85c0e17b34042dd5ab9/src/unix_term.rs#L92

atanunq:
Fixes #66

The issue that this PR tackles is that if a user presses Esc, no key will be detected until a subsequent key is pressed.
To fix the blocking on escape issue, I use `libc::poll` to check whether there is something to be read from the fd. If not, return the key immediately. If yes, continue reading bytes one by one until we find something familiar. `read_single_key` will block if there is nothing to be read at all.
To achieve this I created a `read_single_char` function that will immediately read, or return None if the fd is not ready.

An issue in the implementation was that when Esc is pressed + some unrecognized character, only Esc was returned and the next chars were lost. I added an additional Key enum, `UnknownEscSeq`, that can holds the consumed characters.

##### On key recognition:

The amount of supported keys could be greatly extended. I think, however, that it should be in a different PR. The list is huge - [spec](https://www.xfree86.org/current/ctlseqs.html) + [how rustyline does it](https://github.com/kkawakam/rustyline/blob/master/src/tty/unix.rs). In order to attempt something like this, it would require some refactoring. For instance, make the Key enum a struct to hold a Modifiers enum. Also, the Windows implementation will have to be kept in sync.

This PR adds the logic to detect a few sequences in `read_single_key` but this should be a temporary solution. Since the logic is not too complex for now, I decided to keep it there. It should be extracted once CSI and SS3 sequences are properly handled.

On the other hand, rustyline might be introduced and make all of this unnecessary.

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -81,32 +81,19 @@ pub fn read_secure() -> io::Result<String> {
 }
 
 fn read_single_byte(fd: i32) -> io::Result<Option<u8>> {
-    let mut readfds = core::mem::MaybeUninit::uninit();
-    unsafe {
-        libc::FD_SET(fd, readfds.as_mut_ptr());
-    };
-    let mut readfds = unsafe { readfds.assume_init() };
-
-    // zero timeout, i.e check if there is something to be read right now
-    let mut timeout = libc::timeval {
-        tv_sec: 0,
-        tv_usec: 0,
+    let mut pollfd = libc::pollfd {
+        fd,
+        events: libc::POLLIN,
+        revents: 0,
     };
 
-    let ret = unsafe {
-        libc::select(
-            fd + 1,
-            &mut readfds as *mut _,
-            std::ptr::null_mut(),
-            std::ptr::null_mut(),
-            &mut timeout,
-        )
-    };
+    // timeout of zero means that it will not block
+    let ret = unsafe { libc::poll(&mut pollfd as *mut _, 1, 0) };
     if ret < 0 {
         return Err(io::Error::last_os_error());
     }
 
-    let is_ready = unsafe { libc::FD_ISSET(fd, &mut readfds as *mut _) };
+    let is_ready = pollfd.revents & libc::POLLIN != 0;
 
     if is_ready {
         //there is something to be read
@@ -243,23 +230,14 @@ pub fn read_single_key() -> io::Result<Key> {
         None => {
             //there is no subsequent byte ready to be read, block and wait for input
 
-            let mut readfds = core::mem::MaybeUninit::uninit();
-            unsafe {
-                libc::FD_SET(fd, readfds.as_mut_ptr());
+            let mut pollfd = libc::pollfd {
+                fd,
+                events: libc::POLLIN,
+                revents: 0,
             };
-            let mut readfds = unsafe { readfds.assume_init() };
 
-            // block until there is something to be read
-            let ret = unsafe {
-                libc::select(
-                    fd + 1,
-                    &mut readfds as *mut _,
-                    std::ptr::null_mut(),
-                    std::ptr::null_mut(),
-                    // null timeout pointer means that it will block indefinitely
-                    std::ptr::null_mut(),
-                )
-            };
+            // negative timeout means that it will block indefinitely
+            let ret = unsafe { libc::poll(&mut pollfd as *mut _, 1, -1) };
             if ret < 0 {
                 return Err(io::Error::last_os_error());
             }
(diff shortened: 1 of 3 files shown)


src/unix_term.rs
```diff
@@ -90,76 +126,123 @@ pub fn read_single_key() -> io::Result<Key> {
...
-                    "read interrupted",
-                ))
+        }
+        Some(c) => {
+            let byte = c as u8;
+            let mut buf: [u8; 4] = [byte, 0, 0, 0];
+
+            if byte & 224u8 == 192u8 {
+                // a two byte unicode character
+                let read = unsafe { libc::read(fd, buf[1..].as_mut_ptr() as *mut libc::c_void, 1) };
```
pksunkara, line 189: Instead of this duplication of code, let's move this reading logic to a different function and use it here and in `read_char`.

src/unix_term.rs
```diff
@@ -90,76 +126,123 @@ pub fn read_single_key() -> io::Result<Key> {
...
+                    Err(io::Error::new(
+                        io::ErrorKind::Interrupted,
+                        "read interrupted",
+                    ))
+                } else {
+                    Ok(key_from_escape_codes(&buf[..2 as usize]))
+                }
+            } else if byte & 240u8 == 224u8 {
+                // a three byte unicode character
+                let read = unsafe { libc::read(fd, buf[1..].as_mut_ptr() as *mut libc::c_void, 2) };
```
pksunkara, line 202: Same comment as above.

src/unix_term.rs
```diff
@@ -90,76 +126,123 @@ pub fn read_single_key() -> io::Result<Key> {
...
+                    Err(io::Error::new(
+                        io::ErrorKind::Interrupted,
+                        "read interrupted",
+                    ))
+                } else {
+                    Ok(key_from_escape_codes(&buf[..3 as usize]))
+                }
+            } else if byte & 248u8 == 240u8 {
+                // a four byte unicode character
+                let read = unsafe { libc::read(fd, buf[1..].as_mut_ptr() as *mut libc::c_void, 3) };
```
pksunkara, line 215: Same comment as above.

src/unix_term.rs
```diff
@@ -90,76 +126,123 @@ pub fn read_single_key() -> io::Result<Key> {
...
+                let read = unsafe { libc::read(fd, buf[1..].as_mut_ptr() as *mut libc::c_void, 1) };
+                if read < 0 {
+                    Err(io::Error::last_os_error())
+                } else if buf[1] == b'\x03' {
+                    Err(io::Error::new(
+                        io::ErrorKind::Interrupted,
+                        "read interrupted",
+                    ))
+                } else {
+                    Ok(key_from_escape_codes(&buf[..2 as usize]))
```
pksunkara, line 198: Do we need to use this function anymore?

pksunkara (changes requested):
Great work!!!

I was thinking about using rustyline here but they didn't have a public read keys feature. It's all about reading a line and different modes of it.

If we can port the logic to this crate, we can maybe use this crate in rustyline. (Probably won't happen)

src/unix_term.rs
atanunq: Very good point, created `read_bytes` that is similar to `libc::read` and does the error handling.

src/unix_term.rs
atanunq: I removed the character recognition from it. However, I had to rename it to `key_from_utf8` and use it as a helper method when dealing with multibyte chars.

pksunkara (approved):

src/unix_term.rs
```diff
@@ -176,28 +250,13 @@ pub fn read_single_key() -> io::Result<Key> {
     rv
 }
 
-pub fn key_from_escape_codes(buf: &[u8]) -> Key {
```
pksunkara, line 179: @atanunq Do you think this key_from_escape_code is still needed? What if the user has characters already and just wants to check instead of reading from us?

atanunq: I am not convinced this function was accessible outside the crate.
You have a point, though. We should probably move it out of `unix_term.rs` if we are going to do that.

pksunkara: It's not high priority. Just something I wanted to note down.

PR #85, opened 2021-01-23
feat: unix term add Control-A (home), Control-E (end) and Control-H support

ttys3:
feat: unix term add Control-A (home), Control-E (end) and Control-H (8) support

ref to https://github.com/prompt-toolkit/python-prompt-toolkit/blob/0ae9aa688e0b3f81a29a33015ca1f8a9adb05c6c/prompt_toolkit/input/ansi_escape_sequences.py#L26

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -214,6 +214,9 @@ pub fn read_single_key() -> io::Result<Key> {
                     '\n' | '\r' => Key::Enter,
                     '\x7f' => Key::Backspace,
                     '\t' => Key::Tab,
+                    '\x01' => Key::Home,      // Control-A (home)
+                    '\x05' => Key::End,       // Control-E (end)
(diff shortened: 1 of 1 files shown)


pksunkara:
Aren't they already supported https://github.com/mitsuhiko/console/blob/cc3fcc154a993501222b0ddf527d5f055d0329b7/src/unix_term.rs#L160-L162?

pksunkara (changes requested):
Commented

ttys3:
@pksunkara  no,  `^[[H` is not `^A`,  `^[[F` is also not `^E`, the same doese Control-H

I tested this once again.

in the other way, if it is the same, the python `prompt_toolkit` lib would not impiment it twice ? isn't it?

ttys3:
and, use Control-A (home), Control-E (end) and Control-H will keep your hands in the main keyboard area, without move to the right far away

pksunkara (approved):

PR #93, opened 2021-02-10
Allow creating a Term from an arbitrary Read/Write pair.

stuhood:
Fixes https://github.com/mitsuhiko/console/issues/34. Method and struct names are eminently bikesheddable.

Unfortunately there are some public API changes necessary: described inline.

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -198,7 +198,7 @@ impl Term {
         match self.inner.target {
             TermTarget::Stderr => Style::new().for_stderr(),
             TermTarget::Stdout => Style::new().for_stdout(),
-            // TODO: Needs an implementation. But also: not Copy.
+            // TODO: Needs an implementation. But also: is not Copy.
             #[cfg(unix)]
             TermTarget::ReadWritePair(_) => Style::new().for_stderr(),
         }
--- a/src/term.rs
+++ b/src/term.rs
@@ -10,11 +10,30 @@ use std::os::windows::io::{AsRawHandle, RawHandle};
 
 use crate::{kb::Key, utils::Style};
 
+#[cfg(unix)]
+trait TermWrite: Write + std::fmt::Debug + AsRawFd + Send {}
+#[cfg(unix)]
+impl<T: Write + std::fmt::Debug + AsRawFd + Send> TermWrite for T {}
+
+#[cfg(unix)]
+trait TermRead: io::Read + std::fmt::Debug + AsRawFd + Send {}
+#[cfg(unix)]
+impl<T: io::Read + std::fmt::Debug + AsRawFd + Send> TermRead for T {}
+
+#[cfg(unix)]
+#[derive(Debug, Clone)]
+pub struct ReadWritePair {
+    read: Arc<Mutex<dyn TermRead>>,
+    write: Arc<Mutex<dyn TermWrite>>,
+}
+
 /// Where the term is writing.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 pub enum TermTarget {
     Stdout,
     Stderr,
+    #[cfg(unix)]
+    ReadWritePair(ReadWritePair),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/term.rs
```diff
@@ -10,11 +10,24 @@ use std::os::windows::io::{AsRawHandle, RawHandle};
...
+
+#[derive(Debug, Clone)]
+pub struct ReadWritePair {
+    read: Arc<Mutex<dyn TermRead>>,
+    write: Arc<Mutex<dyn TermWrite>>,
+}
+
 /// Where the term is writing.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
```
stuhood, line 26: The derivation of `Copy` just isn't possible in a safe way for arbitrary `Read`/`Write` destinations. And allowing for equality checks is of questionable value, although we can add it back if need be.

src/term.rs
```diff
@@ -157,19 +170,36 @@ impl Term {
...
+
     /// Returns the style for the term
     #[inline]
     pub fn style(&self) -> Style {
-        match self.target() {
+        match self.inner.target {
             TermTarget::Stderr => Style::new().for_stderr(),
             TermTarget::Stdout => Style::new().for_stdout(),
+            // TODO: Needs an implementation. But also: not Copy.
+            TermTarget::ReadWritePair(_) => Style::new().for_stderr(),
```
stuhood, line 195: `Style` is not `Copy`, so this method is creating a new one for each call. Should probably return a reference.

Additionally, there is a question of whether someone constructing an arbitrary `Term` would want to be able to `Style` it explicitly.

stuhood:
The `ubuntu-latest, i686-unknown-linux-gnu, 1.40.0` failure looks like a flake, but otherwise this should be reviewable.

src/term.rs
```diff
@@ -157,19 +176,38 @@ impl Term {
...
     /// Returns the style for the term
     #[inline]
     pub fn style(&self) -> Style {
-        match self.target() {
+        match self.inner.target {
             TermTarget::Stderr => Style::new().for_stderr(),
             TermTarget::Stdout => Style::new().for_stdout(),
+            // TODO: Needs an implementation. But also: is not Copy.
+            #[cfg(unix)]
+            TermTarget::ReadWritePair(_) => Style::new().for_stderr(),
```
pksunkara, line 203: I don't think this is good. What if people want to emulate stdout?

pksunkara (changes requested):

src/term.rs
pksunkara: @stuhood Any update on this PR?

stuhood: Sorry for the delay: will try to get back to this this week.

stuhood: @pksunkara : Would you be ok with changing the signature of this method to return `&Style` so that we can avoid creating a new `Style` instance each time it is called? If so, I think that we can make the `Style` configurable in the constructor.

pksunkara: Hmm, no. I think the pattern people use is that the style is movable.

stuhood: `Style` does not currently derive `Copy`, so I don't think that that is the case... I could derive `Copy` there, and then each call to this method would clone the `Style`: is that preferable?

pksunkara: Yup.

stuhood: Added an optional `Style`.

stuhood:
@pksunkara : Added an optional `Style`.

Also, the names of structs/methods continue to be completely open to change if you have any thoughts there.

src/term.rs
```diff
@@ -10,11 +10,31 @@ use std::os::windows::io::{AsRawHandle, RawHandle};
 
 use crate::{kb::Key, utils::Style};
 
+#[cfg(unix)]
+trait TermWrite: Write + std::fmt::Debug + AsRawFd + Send {}
+#[cfg(unix)]
+impl<T: Write + std::fmt::Debug + AsRawFd + Send> TermWrite for T {}
+
+#[cfg(unix)]
+trait TermRead: io::Read + std::fmt::Debug + AsRawFd + Send {}
```
pksunkara, line 19: Let's import `Read` and `Debug` at the start.

src/term.rs
pksunkara: I would add it back to maintain as much backward compatibility as possible.

src/term.rs
```diff
@@ -157,19 +177,48 @@ impl Term {
...
+            TermTarget::ReadWritePair(ReadWritePair { ref style, .. }) => style.clone(),
         }
     }
 
     /// Returns the target
     #[inline]
-    pub fn target(&self) -> TermTarget {
-        self.inner.target
+    pub fn target(&self) -> &TermTarget {
+        &self.inner.target
```
pksunkara, line 221: I dont think we can clone here, right?

stuhood: We can clone here... I'll swap it back to preserve backwards compatibility.

In general though: I think that it would be more idiomatic moving forward to use references in more cases, and then letting the caller clone if they need to.

src/term.rs
stuhood: `Copy` cannot be added back, since no useful file handles are `Copy` (without `unsafe`).

`Eq`/`PartialEq` might be possible to add back, but it would need a strange/custom implementation (ie, assigning the `Term` a unique id?), because neither of those traits is object safe:
```
error[E0038]: the trait `TermWrite` cannot be made into an object
  --> src/term.rs:28:22
   |
28 |     write: Arc<Mutex<dyn TermWrite>>,
   |                      ^^^^^^^^^^^^^ `TermWrite` cannot be made into an object
   |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
  --> src/term.rs:15:44
   |
15 | trait TermWrite: Write + Debug + AsRawFd + PartialEq + Send {}
   |       ---------                            ^^^^^^^^^ ...because it uses `Self` as a type parameter
   |       |
   |       this trait cannot be made into an object...
```

pksunkara (approved):

PR #96, opened 2021-04-05
add shift to recognized inputs

Archina:
This fixes an [issue](https://github.com/mitsuhiko/dialoguer/issues/110) on a crate that depends on this one, namely [dialoguer](https://github.com/mitsuhiko/dialoguer).

The problem is that shift is not recognized as a valid input under windows cancelling the text_input routine. This allows `shift` to be handled and with that the input of upper case letters on Windows 10.

### DIFF
--- a/src/kb.rs
+++ b/src/kb.rs
@@ -20,6 +20,7 @@ pub enum Key {
     Tab,
     BackTab,
     Del,
+    Shift,
     Insert,
     PageUp,
     PageDown,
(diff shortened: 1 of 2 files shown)


Archina:
White this work as a workaround - the real issue is that dialoguer depends on checking against Key::Unknown for an erroneous use case. I think it would be a better idea to change `read_key` in the `term.rs` file to return an io::Result specifying the inability to connect to the terminal `io::ErrorKind::NotConnected`, instead of returning `Key::Unknown` - to me this appears to be misleading.

pksunkara (approved):

PR #97, opened 2021-04-15
Exported `pad-str-with` function

lebensterben:
Hi, I think `pad-str-with` is not a implementation detail and should be available to uses.


pksunkara (approved):

pksunkara:
Format check failing

lebensterben:
Fixed with `cargo fmt`.

Issue #102, opened 2021-08-13
Text input yields Key::Unknown on windows_term using alt 


Archina:
This relates to: https://github.com/mitsuhiko/dialoguer/issues/116

Dialoguer listens to Key::Unknown events to cancel the UserInput[[1](https://github.com/mitsuhiko/dialoguer/blob/8762a2e7c1ae7fe0db9eb8ed75a2c46a73f653b4/src/prompts/input.rs#L351)]. This works fine on Linux. On Windows however, windows_term generates Key::Unknown events for special characters.

On my English keyboard layout I can use the ¿ (alt+/) character to create these events. This results in the whole text_input crashing for dialoguer.

Originally I planned on just adding `Key::Alt` to the list. However winapi also seems to identify the key wrong (the events comes in as 0x12) which would reference the menu key[[2](https://docs.rs/winapi/0.3.9/winapi/um/winuser/constant.VK_MENU.html)]. Which should be incorrect.

I could map the events to `Key::Composite` just to mark that a key composition event is taking place or use `Key::Alt` and comment the location to add a fix if/when winapi is fixed/changed.

PR #103, opened 2021-08-13
fixes windows composite key: #102 - add composite key

Archina:
[#102 - Text input yields Key::Unknown on windows_term using alt](https://github.com/mitsuhiko/console/issues/102)

### DIFF
--- a/src/kb.rs
+++ b/src/kb.rs
@@ -19,7 +19,7 @@ pub enum Key {
     End,
     Tab,
     BackTab,
-    Composite,
+    Alt,
     Del,
     Shift,
     Insert,
--- a/src/kb.rs
+++ b/src/kb.rs
@@ -19,6 +19,7 @@ pub enum Key {
     End,
     Tab,
     BackTab,
+    Composite,
     Del,
     Shift,
     Insert,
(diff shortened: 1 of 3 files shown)


pksunkara (approved):

pksunkara:
Okay, what is a composite key and how is it related to macos and ubuntu?

Archina:
As stated in the linked issue, the composite key does not really exist. It also does not relate to linux. The mapping is more for the Alt key - however, calling it Key::Alt seemed misleading as the mapping within winapi seems to be incorrect as well (it points to the Menu key instead).

The issue is that the windows term generates Key::Unknown + Key::Char events when using this technique. (You should be able to reproduce this by using altGr + / to create a ¿ character). This causes dialoguer to crash.

pksunkara:
If you check https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes, VK_MENU **is** ALT key. So, instead of composite, let's put `Key::Alt`

Archina:
Then I misunderstood that. I will use Key::Alt then instead. Thank you very much!

pksunkara (approved):

PR #119, opened 2022-04-29
Further documentation improvements

grunweg:
I made a few further improvements to the doc comments, similar in spirit to #110.

I don't know how to get rid of the merge commit (github added that for me). Feel free to rebase for me and remove it.

PR #120, opened 2022-04-29
More documentation improvements: clarifying behaviour

grunweg:
In addition to some minor stylistic edits (see inline), I
- clarified how the cursor is positioned after the various clear functions
- clarify that indices in move_to are 0-based. (I was wondering about that; it seems harmless to guarantee this since the ANSI code move-to sequences that this is based on have the same semantics.)
- document how move_cursor_{left,right,up,down} behave when called with a larger value than possible.

Based on #119; only the last four commits are new.


pksunkara (approved):

grunweg:
Thanks for the fast review!

PR #133, opened 2022-09-19
Upgrade to terminal_size v0.2

benesch:


pksunkara:
That version uses different dependencies to libc and winapi. So, doing this now without changing our own dependencies just adds more dependencies to compile time for not much benefit.

benesch:
Will you accept a revision that does the version upgrade of terminal_size
and also ports console over to rustix and windows-sys? If so please reopen.

benesch:
Thanks! I'll take a look at this when I can.

benesch:
Ah, never mind. I'm afraid I just won't have the time.

PR #145, opened 2022-11-14
Implement `clear_to_end_of_line`

hampuslidin:


JeromeSchmied:
I'd love to see this merged, as I've started experimenting with a `console` backend for `ratatui`: https://github.com/ratatui/ratatui/pull/1818, and this is kind of a blocking issue

djc:
This looks small enough that it seems acceptable if rebased.

PR #151, opened 2023-01-02
Fix invalid terminal size being returned on Windows

### LINKED ISSUE #150: Invalid terminal size returned on Windows
The documentation of [`Term::size`](https://docs.rs/console/latest/console/struct.Term.html#method.size) states: 
"Return the terminal size in rows and columns ..."

However on Windows it currently returns the width first, then the height:
https://github.com/console-rs/indicatif/blob/main/src/draw_target.rs#L113

This regression was introduced during the migration to the windows crate: https://github.com/console-rs/console/commit/7febd3c96adbedaa29a0fa8edf9c3c1ee850570a#
Currently, this causes wide progress bars not render properly: https://github.com/console-rs/indicatif/issues/505

AronParker:
Fixes #150

### DIFF
--- a/src/windows_term/mod.rs
+++ b/src/windows_term/mod.rs
@@ -137,9 +137,10 @@ pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
         return None;
     }
 
-    let w = (csbi.srWindow.Right - csbi.srWindow.Left + 1) as u16;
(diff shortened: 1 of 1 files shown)


Araxeus:
Thank you!  it also broke scrolling in [dialoguer](https://crates.io/crates/dialoguer)

glad to see this finally fixed

PR #181, opened 2023-08-09
Update read_line behaviour to read_line_initial_text

thibmeu:
`term.read_line()` is not using the declared term to request the input. It relies on stdin. This is not the case of `term.read_line_initial_text("")`, which reads key one by one using `self.read_key()`.

This commits corrects this behaviour, and removes divergence between the two methods.

If `read_line` and `read_line_initial_text` are meant to have a different behaviour, their documentation needs to be updated. I'd be glad to update this PR if that's the case.


thibmeu:
@mitsuhiko tagging you for review as you've been the last one to merge changes.

PR #186, opened 2023-09-24
Use `out.as_raw_fd()` to get size in `unix_term`

### LINKED ISSUE #185: `unix_term::terminal_size` uses `STDIN_FILENO` for `isatty` instead of configured output
For example;

https://github.com/console-rs/console/blob/0567bdcb6a523466586dc9ff0692c2e776f20a5f/src/unix_term.rs#L49-L51

Should likely use `out.as_raw_fd()` instead. I *think* the same issue might exist in a few other places in that file that don't currently take a `Term` at all.

tgolsson:
Fixes #185. Tested as a resolution to https://github.com/pantsbuild/pants/pull/19931#issuecomment-1732652244 where the line is always truncated at 80 chars.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Bugfixes
+
+* Properly use configured output of `Term` to get terminal size (#186)
+
 ## 0.15.7
 
 ### Enhancements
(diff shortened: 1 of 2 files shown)


jgrund:
Ping, any updates on this? We would like to rely on this change for a downstream crate

PR #195, opened 2023-12-31
support for synchronized output

Funami580:
I got the advice that support for synchronized output should be implemented in this crate instead of indicatif (related indicatif PR: https://github.com/console-rs/indicatif/pull/618).

Currently, only checking whether synchronized output is supported is implemented. Before I continue, I want to know what the public API should look like (previous examples: https://github.com/console-rs/indicatif/pull/618#issuecomment-1872589850).

@djc already suggested to use guards, and that I should ask the console maintainers, too. I'd also like to see some vague code for the public API, for example:
```rs
let term = Term::stdout();
if term.supports_sync_update() {
	term.begin_sync_update();
}
// do something
if term.supports_sync_update() {
	term.end_sync_update();
}
term.flush();
```

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -5,7 +5,8 @@ use std::io;
 use std::io::{BufRead, BufReader};
 use std::mem;
 use std::os::unix::io::AsRawFd;
-use std::ptr;
+use std::os::unix::io::FromRawFd;
+use std::os::unix::io::IntoRawFd;
 use std::str;
 
 use crate::kb::Key;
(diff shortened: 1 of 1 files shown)


src/unix_term.rs
```diff
@@ -360,3 +361,289 @@ pub fn wants_emoji() -> bool {
...
+    impl ResponseParser {
+        const fn new() -> Self {
+            Self {
+                state: ResponseParserState::None,
+                response: u8::MAX,
+            }
+        }
+
+        fn process_byte(&mut self, byte: u8) {
+            match byte {
```
djc, line 488: Suggest writing this as such:

```rust
self.state = match (self.state, byte) {
    ..
};
```

src/unix_term.rs
```diff
@@ -360,3 +361,289 @@ pub fn wants_emoji() -> bool {
 pub fn set_title<T: Display>(title: T) {
     print!("\x1b]0;{}\x07", title);
 }
+
+fn with_raw_terminal<R>(f: impl FnOnce(&mut fs::File) -> R) -> io::Result<R> {
```
djc, line 365: If this is only used within `sync_output`, I would move it inside the module.

src/unix_term.rs
```diff
@@ -360,3 +361,289 @@ pub fn wants_emoji() -> bool {
...
+
+    Ok(result)
+}
+
+pub fn supports_synchronized_output() -> bool {
+    *sync_output::SUPPORTS_SYNCHRONIZED_OUTPUT
+}
+
+/// Specification: https://gist.github.com/christianparpart/d8a62cc1ab659194337d73e399004036
+mod sync_output {
```
djc, line 439: I think Rust style would suggest not abbreviating `synchronized_` to `sync_` for the module name.

djc (commented):
Some stylistic nits.

PR #206, opened 2024-02-07
Add utility function `ansi::slice_ansi_str`

remi-dupre:
Hi!

While working on a bug of `indicatif` (cf. https://github.com/console-rs/indicatif/pull/627), I needed an ANSI-aware slicing method. I think it made more sense to add it in this crate.

```rust
/// Slice a `&str` in terms of text width. This means that only the text
/// columns strictly between `start` and `stop` will be kept.
///
/// If a multi-columns character overlaps with the end of the interval it will
/// not be included. In such a case, the result will be less than `end - start`
/// columns wide.
///
/// If non-empty head and tail are specified, they are inserted between the
/// ANSI symbols from truncated bounds and the slice.
///
/// This ensures that escape codes are not screwed up in the process.
pub fn slice_str<'a>(s: &'a str, head: &str, bounds: Range<usize>, tail: &str) -> Cow<'a, str> {
    ...
}
```

This is a generalisation of `truncate_str` 😃

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -1,4 +1,5 @@
 use alloc::borrow::Cow;
+use core::ops::Range;
 use core::{
     fmt::{self, Debug, Formatter},
     sync::atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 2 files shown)


src/unix_term.rs
```diff
@@ -56,7 +56,7 @@ pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
         #[allow(clippy::useless_conversion)]
         libc::ioctl(out.as_raw_fd(), libc::TIOCGWINSZ.into(), &mut winsize);
         if winsize.ws_row > 0 && winsize.ws_col > 0 {
-            Some((winsize.ws_row as u16, winsize.ws_col as u16))
+            Some((winsize.ws_row, winsize.ws_col))
```
remi-dupre, line 59: This was just a random (new?) clippy lint : these are already u16 :shrug:

src/utils.rs
```diff
@@ -71,7 +66,17 @@ pub fn set_colors_enabled_stderr(val: bool) {
 
 /// Measure the width of a string in terminal characters.
 pub fn measure_text_width(s: &str) -> usize {
-    str_width(&strip_ansi_codes(s))
+    #[cfg(feature = "ansi-parsing")]
+    {
+        AnsiCodeIterator::new(s)
+            .filter(|(_, is_ansi)| !is_ansi)
+            .map(|(sub, _)| str_width(sub))
+            .sum()
```
remi-dupre, line 74: Not related to the MR, but while implementing a less optimized draft of my fix I thought this could spare a few allocs. This is probably not something to care about so I wouldn't mind if you want to revert this for the sake of simplification.

djc:
Can you avoid `let .. else` in order to pass the MSRV tests?

remi-dupre:
Indeed, that should be good now :+1:

remi-dupre:
And format is also a recent-is feature, on it :grimacing: 

I'll try to build on 1.56, it didn't work last time I tried it :disappointed:

remi-dupre:
Should be good, `make test` runs with rust 1.56 on my machine :+1:

src/unix_term.rs
```diff
@@ -59,7 +59,7 @@ pub(crate) fn terminal_size(out: &Term) -> Option<(u16, u16)> {
         winsize
     };
     if winsize.ws_row > 0 && winsize.ws_col > 0 {
-        Some((winsize.ws_row as u16, winsize.ws_col as u16))
+        Some((winsize.ws_row, winsize.ws_col))
```
remi-dupre, line 62: This is a random new clippy lint

remi-dupre:
Hi!

I bumped into this while doing some cleanup and I think it might still be useful. So I've rebased it on latest main!

djc:
I definitely don't want to accept a semver-incompatible change for this. Is there a way to make this compatible?

remi-dupre:
Hi!

You're right that it would be a shame.

Although I think I made a mistake with semver : I only added a new function without affecting the behavior of the existing, as we're still on unstable scheme *0.x.y* this should only affect the last component.

I changed the target version to *0.15.12* and made sure that I didn't modify previously existing test cases :+1:

djc:
Okay, this doesn't currently pass the MSRV jobs in CI, so that's something that will have to be fixed. I'd like a cleaner set of changes for review, with separate commits or PRs for each of these changes:

- clippy fixes
- saving allocations on `measure_text_width()`
- introducing `slice_ansi_str()`

Please avoid unrelated changes as much as possible (as such, keep functions/logic in the place as much as possible to minimize the diffs). You can leave the version bumping and changelog to me (note that the `CHANGELOG.md` file is no longer being used).

remi-dupre:
Hello mon ami ! 🥖

Those are super relevant comments :+1: 

1. I've created two smaller MRs for parasitic changes
  a. https://github.com/console-rs/console/pull/247
  b. https://github.com/console-rs/console/pull/246
2. Following your advice, I've created another to suggest a fix to the MSRV issue : https://github.com/console-rs/console/pull/248
3. I've removed version & changelog related diffs

remi-dupre:
I'm putting this in draft until #250 is solved.

src/utils.rs
```diff
@@ -807,77 +808,124 @@ pub(crate) fn char_width(_c: char) -> usize {
...
+        let mut pos = 0; // Current search index by width
+        let mut code_iter = AnsiCodeIterator::new(s).peekable();
+
+        // Search for the begining of the slice while collecting heading ANSI
+        // codes
+        let mut front_ansi = String::new(); // ANSI codes found before bound start
+        let mut slice_start = 0; // Current search index by bytes
+
+        // Extract the leading slice, which *may be mutated* to remove just its first character.
+        'search_slice_start: while pos < bounds.start {
```
djc, line 833: I'm counting 5 loops here, two of which are nested. Fundamentally, why is that necessary? It seems like there is more complexity here than is warranted.

remi-dupre:
Hi!

I'm a process of cleaning up: I've rebased this PR and will be available for enhancement if you feel this change is still relevant

src/utils.rs
djc: But you are now allocating for `front_ansi` which presumbly isn't great for performance...

src/utils.rs
```diff
@@ -890,78 +891,124 @@ pub(crate) fn char_width(_c: char) -> usize {
     1
 }
 
-/// Truncates a string to a certain number of characters.
+/// Slice a `&str` in terms of text width. This means that only the text
+/// columns strictly between `start` and `stop` will be kept.
```
djc, line 895: It's not `stop` but `end` presumably.

I think this docstring could use a few examples to make the expected behavior more clear.

djc (commented):
I think it is still relevant, but it's also still quite complex.

Apart from my other comments, I'd suggest splitting the ANSI-aware and non-ANSI-aware logic into two different function definitions to save one level of indentation.

PR #210, opened 2024-04-04
Moved from unicode-width to unicode-display-width for visual grapheme width estimation

Silvea12:
This change had to be made in both crates in order to be updated correctly.
The console-rs crate has been bumped an extra version as a result of a behavior change of `char_width()` with the `unicode-display-width` feature enabled - it now returns always 1 or 2, never 0 as it used to.
The indicatif crate may also need to be bumped a full version as a result of the dependency change, but I'm not sure if you'd consider this a breaking change to the same degree, as it's not an advertised feature.

Related issue: https://github.com/console-rs/indicatif/issues/638
Related PR: https://github.com/console-rs/indicatif/pull/639

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -720,11 +720,27 @@ fn str_width(s: &str) -> usize {
 #[cfg(feature = "ansi-parsing")]
 fn char_width(c: char) -> usize {
     #[cfg(feature = "unicode-display-width")]
-    {
-        match unicode_display_width::is_double_width(c) {
-            true => 2,
-            false => 1,
+    if c < '\u{7F}' {
+        if c >= '\u{20}' {
+            // U+0020 to U+007F (exclusive) are single-width ASCII codepoints
+            1
+        } else if c == '\0' {
+            // U+0000 *is* a control code, but it's special-cased
+            0
+        } else {
+            // U+0001 to U+0020 (exclusive) are control codes
+            0
         }
+    } else if c >= '\u{A0}' {
+        // No characters >= U+00A0 are control codes, so we can consult the lookup tables
+        if unicode_display_width::is_double_width(c) {
+            2
+        } else {
+            1
+        }
+    } else {
+        // U+007F to U+00A0 (exclusive) are control codes
+        0
     }
     #[cfg(not(feature = "unicode-display-width"))]
     {
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -707,12 +707,11 @@ impl<'a, 'b> fmt::Display for Emoji<'a, 'b> {
 }
 
 fn str_width(s: &str) -> usize {
-    #[cfg(feature = "unicode-width")]
+    #[cfg(feature = "unicode-display-width")]
     {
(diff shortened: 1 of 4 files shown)


Silvea12:
Pulling in @djc as requested

src/utils.rs
```diff
@@ -707,25 +707,26 @@ impl<'a, 'b> fmt::Display for Emoji<'a, 'b> {
 }
 
 fn str_width(s: &str) -> usize {
-    #[cfg(feature = "unicode-width")]
+    #[cfg(feature = "unicode-display-width")]
     {
-        use unicode_width::UnicodeWidthStr;
-        s.width()
+        unicode_display_width::width(s) as usize
```
Silvea12, line 712: Note: this usize cast is here to avoid changing APIs. It may not be what is wanted?

src/utils.rs
```diff
@@ -707,25 +707,26 @@ impl<'a, 'b> fmt::Display for Emoji<'a, 'b> {
...
 }
 
 #[cfg(feature = "ansi-parsing")]
 fn char_width(c: char) -> usize {
-    #[cfg(feature = "unicode-width")]
+    #[cfg(feature = "unicode-display-width")]
     {
-        use unicode_width::UnicodeWidthChar;
-        c.width().unwrap_or(0)
+        match unicode_display_width::is_double_width(c) {
```
djc, line 724: For compatibility, maybe we should replicate the logic from `unicode-width` here for control characters? Maybe like:

```rust
match c.is_control() {
     true if c == '\x00' => ?,
     true => 0,
     false => match unicode_display_width::is_double_width(c) {
         true => 2,
         false => 1,
     }
}
```

src/utils.rs
djc: This seems okay to me. Would sure be nice to avoid breaking compatibility for this.

Silvea12:
Worth an extra note: I'm not sure if `truncate_str()` needs to be adjusted - it'll depend on how ANSI is handled here, and how much we need to care about grapheme clusters. I'm not 100% sure if we need to change `if unicode_display_width::is_double_width(c) {` into `if c == '\u{FEFF}' || unicode_display_width::is_double_width(c) {` or not, as a result.

src/utils.rs
```diff
@@ -707,25 +707,42 @@ impl<'a, 'b> fmt::Display for Emoji<'a, 'b> {
...
 }
 
 #[cfg(feature = "ansi-parsing")]
 fn char_width(c: char) -> usize {
-    #[cfg(feature = "unicode-width")]
-    {
-        use unicode_width::UnicodeWidthChar;
-        c.width().unwrap_or(0)
+    #[cfg(feature = "unicode-display-width")]
+    if c < '\u{7F}' {
```
djc, line 723: Why did you go this route? I'd like to avoid referring to explicit code points as much as possible as it would be likely for these to fall out of sync for Unicode updates -- that's what we're trying to rely on other crates for.

PR #213, opened 2024-06-15
Fix some typos

waywardmonkeys:

PR #214, opened 2024-06-15
ci: Switch from actions-rs and update checkout

waywardmonkeys:
The actions-rs actions have been unmaintained for a long time.


waywardmonkeys:
The CI issues here (and in #213) are from MSRV being broken by an update to `unicode-width` which no longer works with Rust 1.56.

djc:

PR #215, opened 2024-06-15
ci: Force older version of unicode-width for MSRV

waywardmonkeys:
unicode-width 0.1.13 added code which didn't compile with Rust 1.56


waywardmonkeys:
This should fix the failure seen in #213 and #214.

djc:
Bumped the MSRV and updated unicode-width in #227.

PR #218, opened 2024-08-03
Switch from lazy_static to once_cell

alex:
This will ultimately facilitate migrating to the std solution, once the MSRV is high enough.


tamird:
FWIW this was previously reverted in https://github.com/console-rs/console/pull/137 over MSRV and https://github.com/console-rs/console/issues/134 contains discussion.

djc:

PR #219, opened 2024-08-26
Replace lazy_static with once_cell

tamird:


tamird:
Eh, seems #218 beat me to it. @mitsuhiko could you have a look?

djc:

PR #222, opened 2024-08-26
Don't eagerly close tty fd in read_secure

### LINKED ISSUE #221: On Unix, read_secure_line doesn't work if stdin isn't a TTY
On Unix, read_secure_line first checks whether stdin is a TTY, and if not falls back to /dev/tty.
https://github.com/console-rs/console/blob/de2f15a31a8fef0b0e65ef4bdf92cd03c3061dac/src/unix_term.rs#L72

But when I tried
```rust
use std::error::Error;
use console::Term;

fn main() -> Result<(), Box<dyn Error>> {
    let term = Term::stdout();

    let x = term.read_secure_line()?;
    dbg!(x);
    Ok(())
}
```

with `echo hello-world | target/debug/mytest`
it hangs until I press return, after which it prints
```
Error: Os { code: 9, kind: Uncategorized, message: "Bad file descriptor" }
```

I don't think this should happen?

A bit of printf debugging revealed this line to be the source of the error
https://github.com/console-rs/console/blob/de2f15a31a8fef0b0e65ef4bdf92cd03c3061dac/src/unix_term.rs#L97

procfs reveals that the FD has been closed at that point.

and further debugging reveals that, indeed, someone is doing a little bit of FD closing.

https://github.com/console-rs/console/blob/de2f15a31a8fef0b0e65ef4bdf92cd03c3061dac/src/unix_term.rs#L91

this line moves out of `f_tty`, which closes the `File`.

Noratrieb:
The tcsetattr was failing because the fd was closed in the if statement before, where it was moved out.

i tested locally that this fixes #221

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -64,7 +64,7 @@ pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
 }
 
 pub fn read_secure() -> io::Result<String> {
-    let f_tty;
+    let mut f_tty;
     let fd = unsafe {
         if libc::isatty(libc::STDIN_FILENO) == 1 {
             f_tty = None;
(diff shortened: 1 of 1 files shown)


djc:
Thanks!

PR #223, opened 2024-08-27
Improve type safety, extract identical code

tamird:
This is an alternative to #222.


tamird:
@mitsuhiko would it be possible to get a review here and in #220?

tamird:
@djc could you TAL at this one as well?

src/unix_term.rs
```diff
@@ -48,7 +47,7 @@ pub fn c_result<F: FnOnce() -> libc::c_int>(f: F) -> io::Result<()> {
 
 pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
     unsafe {
-        if libc::isatty(out.as_raw_fd()) != 1 {
+        if !is_a_terminal(out) {
```
djc, line 50: Nit: put this in a separate commit?

src/unix_term.rs
```diff
@@ -18,7 +17,7 @@ pub use crate::common_term::*;
 pub const DEFAULT_WIDTH: u16 = 80;
 
 #[inline]
-pub fn is_a_terminal(out: &Term) -> bool {
+pub fn is_a_terminal<T: AsRawFd>(out: &T) -> bool {
```
djc, line 20: Nit: prefer to spell this as `out: &impl AsRawFd` to avoid the indirection via `T`.

src/unix_term.rs
```diff
@@ -37,7 +36,7 @@ pub fn is_a_color_terminal(out: &Term) -> bool {
     }
 }
 
-pub fn c_result<F: FnOnce() -> libc::c_int>(f: F) -> io::Result<()> {
+fn c_result<F: FnOnce() -> libc::c_int>(f: F) -> io::Result<()> {
```
djc, line 39: Move the visibility changes into a separate commit as well? Ideally enable `warn(unreachable_pub)` so we can get them all at once.

src/unix_term.rs
```diff
@@ -66,41 +65,71 @@ pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
-                .open("/dev/tty")?;
-            let fd = f.as_raw_fd();
-            f_tty = Some(BufReader::new(f));
-            fd
+enum Input<T> {
+    Stdin(io::Stdin),
+    File(T),
+}
+
+impl<T> Input<T>
```
djc, line 73: Can we `impl<T: BufRead> BufRead for Input<T>` instead?

src/unix_term.rs
```diff
@@ -66,41 +65,71 @@ pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
+    if is_a_terminal(&stdin) {
+        Ok(Input::Stdin(stdin))
+    } else {
+        let f = fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .open("/dev/tty")?;
+        Ok(Input::File(f))
+    }
+}
```
djc, line 114: Should make this an `Input` constructor. Suggest using `match` instead of `if` and inlining the `f` binding. Pick good names to distinguish `input()` and `read_secure()`?

djc (commented):
Really like the type safety, would like this cleaned up a bit. Thanks.

src/unix_term.rs
tamird: Done, though note `pub(crate) fn set_title<T: Display>(title: T)` at the bottom of the file.

src/unix_term.rs
tamird: Done.

src/unix_term.rs
tamird: Sure. Done.

src/unix_term.rs
tamird: Done.

src/unix_term.rs
```diff
@@ -66,41 +65,71 @@ pub fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
+
+pub fn read_secure() -> io::Result<String> {
```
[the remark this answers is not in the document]
tamird: Done.

tamird:
Added a commit to clean things up further.

tamird:
Rebased and tested that the keyboard example behaves the same as before on my mac.

src/unix_term.rs
```diff
@@ -66,41 +65,76 @@ pub(crate) fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
+impl Input<BufReader<fs::File>> {
+    fn new() -> io::Result<Self> {
+        Ok(match Input::<fs::File>::new()? {
+            Input::Stdin(s) => Self::Stdin(s),
+            Input::File(f) => Self::File(BufReader::new(f)),
+        })
+    }
+}
+
+impl<T: BufRead> Input<T> {
```
djc, line 106: As mentioned earlier, can we do a full `BufRead` impl instead by just forwarding all the required methods?

src/unix_term.rs
```diff
@@ -66,41 +65,76 @@ pub(crate) fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
-        if libc::isatty(libc::STDIN_FILENO) == 1 {
-            f_tty = None;
-            libc::STDIN_FILENO
+enum Input<T> {
+    Stdin(io::Stdin),
+    File(T),
+}
+
+impl Input<fs::File> {
+    fn new() -> io::Result<Self> {
```
djc, line 74: As suggested in my earlier comment, I think we should give these `new()` methods proper names, maybe `buffered()` and `unbuffered()`?

src/unix_term.rs
```diff
@@ -66,41 +65,76 @@ pub(crate) fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
+            Ok(Input::Stdin(stdin))
         } else {
             let f = fs::OpenOptions::new()
                 .read(true)
                 .write(true)
                 .open("/dev/tty")?;
-            let fd = f.as_raw_fd();
-            f_tty = Some(BufReader::new(f));
-            fd
+            Ok(Input::File(f))
```
djc, line 83: As suggested previously, please inline the `f` binding.

src/unix_term.rs
```diff
@@ -301,27 +335,17 @@ fn read_single_key_impl(fd: i32) -> Result<Key, io::Error> {
...
-                .write(true)
-                .open("/dev/tty")?;
-            tty_f.as_raw_fd()
-        }
-    };
+    let input = Input::<fs::File>::new()?;
+
     let mut termios = core::mem::MaybeUninit::uninit();
-    c_result(|| unsafe { libc::tcgetattr(fd, termios.as_mut_ptr()) })?;
+    c_result(|| unsafe { libc::tcgetattr(input.as_raw_fd(), termios.as_mut_ptr()) })?;
```
djc, line 283: Can we just do `let fd = input.as_raw_fd();` somewhere near the top so we don't have this duplication?

src/unix_term.rs
```diff
@@ -66,41 +65,76 @@ pub(crate) fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
+    fn as_raw_fd(&self) -> RawFd {
+        match self {
+            Self::Stdin(s) => s.as_raw_fd(),
+            Self::File(f) => f.get_ref().as_raw_fd(),
+        }
+    }
+}
+
+pub(crate) fn read_secure() -> io::Result<String> {
+    let mut input = Input::<BufReader<fs::File>>::new()?;
```
djc, line 134: Can we use `let fd = input.as_raw_fd();` here and then reuse the binding?

src/unix_term.rs
```diff
@@ -66,41 +65,76 @@ pub(crate) fn terminal_size(out: &Term) -> Option<(u16, u16)> {
...
-        f.read_line(&mut rv)
-    } else {
-        io::stdin().read_line(&mut rv)
-    };
-
-    c_result(|| unsafe { libc::tcsetattr(fd, libc::TCSAFLUSH, &original) })?;
+    let read_rv = input.read_line(&mut rv);
 
-    // Ensure the fd is only closed after everything has been restored.
-    drop(f_tty);
```
djc, line 103: I think this is a rebase error?

src/unix_term.rs
```diff
@@ -94,6 +94,15 @@ impl Input<BufReader<fs::File>> {
     }
 }
 
+impl<T: Read> Read for Input<T> {
```
djc, line 97: Would prefer to avoid increasing the scope of this PR, please submit this separately after we merge the other changes here?

src/unix_term.rs
tamird: Sure, done. I made them free functions because otherwise we end up with an unconstrained type parameter and would have to write something like `Input::<fs::File>::unbuffered()` which doesn't look very pleasing.

src/unix_term.rs
tamird: Sure, removed.

src/unix_term.rs
tamird: Indeed that was suggested earlier, but I find that harder to read because of the error handling in the `f` expression. I'd prefer not to have `?` in the middle of an expression.

src/unix_term.rs
tamird: `io::Stdin` doesn't implement `BufRead`; its `read_line` method is inherent.

src/unix_term.rs
tamird: Nope. This code was here to ensure that `f_tty` outlives the last use of `fd`, since both represent the same underlying resource whose lifetime is tied to `f_tty`. This is also the reason I won't write `let fd = ...`; calling `as_raw_fd()` is trivial but ensures that the proper lifetimes are preserved.

src/unix_term.rs
tamird: See below. I'd rather not.

src/unix_term.rs
tamird: Same as above.

djc:
Thanks!

PR #226, opened 2024-11-28
unicode-width: 0.1 -> 0.2

nagisa:
indicatif already depends on 0.2, so this helps to deduplicate my dependencies a bit.


djc:

PR #231, opened 2024-12-17
Avoid manual polling

tamird:
Use blocking and non-blocking reads instead.


djc:
Can you add some high-level context/motivation here? Is it feasible to split this in smaller changes (that still compile/pass tests)?

djc:
Can you add some high-level context/motivation here? Is it feasible to split this in smaller changes (that still compile/pass tests)?

tamird:
Updated the commit message.

tamird:
Sorry, didn't address your question. No, I don't think it's feasible to break this into smaller bits.

djc:
Sorry, this is a larger change than I feel comfortable reviewing (since I have limited experience with this crate and the underlying Unix APIs). The transformation from select to temporarily changing the socket to be non-blocking seems pretty tricky to get right and I don't think the motivation is sufficient given the risk.

tamird:
Who would be an appropriate reviewer? @atanunq wrote the original code in https://github.com/console-rs/console/commit/c8ea46fad86246f91c13cd83a9a4ddc935f1b945 and @mitsuhiko modified it in https://github.com/console-rs/console/pull/169.

djc:
Not sure how much capacity @mitsuhiko has for console in the near future.

tamird:
The original code written by @atanunq borrowed from rustyline, which uses polling, but rustyline also uses a non-zero timeout. I'd love for one of these two original authors to weigh in.

Also tagging @pksunkara who approved the original changes.

PR #233, opened 2025-01-22
feat: support gradients

Mathias-Boulay:
As said on [this issue](https://github.com/console-rs/indicatif/issues/691), I said I would propose a PR.

Screenshot From 2025-01-22 15-32-11

# Notable elements and changes

- The API is fully backwards compatible, as `console` was a transitive dependency and I needed my project

- Add supports for multi-step gradients. If the terminal doesn't support "true colors", it will have no effect.

- utils functions to check/force true color support

From a `dotted_string`, you can use `[on_]gradient(_HEX_COLOR)+` 
For example: 
```rust
".magenta.gradient_E50000_FF8D00_FFEEOO_028121_004CFF_770088"
```
This will use a gradient on true colors terminals and fall back on the magenta color for other terminals.

From a `StyledObject<D>` perspective, use `.[on_]gradient(GradientColor)` to add a color to the gradient.

- This PR doesn't bring additional tests, as I want first to be sure the API is final

- The gradient support is heavily inspired by [Tinterm](https://github.com/bratish/Tinterm/)

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -17,10 +17,18 @@ fn default_colors_enabled(out: &Term) -> bool {
         || &env::var("CLICOLOR_FORCE").unwrap_or_else(|_| "0".into()) != "0"
 }
 
+fn default_true_colors_enabled(out: &Term) -> bool {
+    out.features().true_colors_supported()
+}
+
 static STDOUT_COLORS: Lazy<AtomicBool> =
     Lazy::new(|| AtomicBool::new(default_colors_enabled(&Term::stdout())));
+static STDOUT_TRUE_COLORS: Lazy<AtomicBool> =
+    Lazy::new(|| AtomicBool::new(default_true_colors_enabled(&Term::stdout())));
 static STDERR_COLORS: Lazy<AtomicBool> =
     Lazy::new(|| AtomicBool::new(default_colors_enabled(&Term::stderr())));
+static STDERR_TRUE_COLORS: Lazy<AtomicBool> =
+    Lazy::new(|| AtomicBool::new(default_true_colors_enabled(&Term::stderr())));
 
 /// Returns `true` if colors should be enabled for stdout.
 ///
@@ -34,6 +42,12 @@ pub fn colors_enabled() -> bool {
     STDOUT_COLORS.load(Ordering::Relaxed)
 }
 
+/// Returns `true` if true colors should be enabled for stdout.
+#[inline]
+pub fn true_colors_enabled() -> bool {
+    STDERR_TRUE_COLORS.load(Ordering::Relaxed)
+}
+
 /// Forces colorization on or off for stdout.
 ///
 /// This overrides the default for the current process and changes the return value of the
@@ -41,6 +55,15 @@ pub fn colors_enabled() -> bool {
 #[inline]
 pub fn set_colors_enabled(val: bool) {
     STDOUT_COLORS.store(val, Ordering::Relaxed)
+}
+
+/// Forces true colorization on or off for stdout.
+///
+/// This overrides the default for the current process and changes the return value of the
+/// `true_colors_enabled` function.
+#[inline]
+pub fn set_true_colors_enabled(val: bool) {
+    STDOUT_TRUE_COLORS.store(val, Ordering::Relaxed)
 }
 
 /// Returns `true` if colors should be enabled for stderr.
@@ -55,13 +78,28 @@ pub fn colors_enabled_stderr() -> bool {
     STDERR_COLORS.load(Ordering::Relaxed)
 }
 
+/// Returns `true` if true colors should be enabled for stderr.
+#[inline]
+pub fn true_colors_enabled_stderr() -> bool {
+    STDERR_TRUE_COLORS.load(Ordering::Relaxed)
+}
+
 /// Forces colorization on or off for stderr.
 ///
 /// This overrides the default for the current process and changes the return value of the
-/// `colors_enabled` function.
+/// `colors_enabled_stderr` function.
 #[inline]
 pub fn set_colors_enabled_stderr(val: bool) {
     STDERR_COLORS.store(val, Ordering::Relaxed)
+}
+
+/// Forces true colorization on or off for stderr.
+///
+/// This overrides the default for the current process and changes the return value of the
+/// `true_colors_enabled_stderr` function.
+#[inline]
+pub fn set_true_colors_enabled_stderr(val: bool) {
+    STDERR_TRUE_COLORS.store(val, Ordering::Relaxed)
 }
 
 /// Measure the width of a string in terminal characters.
@@ -117,6 +155,45 @@ impl Color {
         match self {
             Color::Color256(_) => true,
             _ => false,
+        }
+    }
+}
+
+/// A Color used in a gradient
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct GradientColor {
+    red: u8,
+    green: u8,
+    blue: u8,
+}
+
+impl GradientColor {
+    fn new(red: u8, green: u8, blue: u8) -> Self {
+        GradientColor {
+            red: red.min(255),
+            green: green.min(255),
+            blue: blue.min(255),
+        }
+    }
+
+    fn from_hex(hex: &str) -> Self {
+        // Remove '#' if present
+        let s = hex.trim_start_matches('#');
+
+        // 6 char hex
+        let mut channels: [u8; 3] = Default::default();
+        for i in (0..5).step_by(2) {
+            channels[i / 2] = u8::from_str_radix(&s[i..i + 2], 16).unwrap_or_default()
+        }
+
+        GradientColor::new(channels[0], channels[1], channels[2])
+    }
+
+    pub(crate) fn interpolate(&self, other: &Self, t: f32) -> Self {
+        GradientColor {
+            red: (self.red as f32 + (other.red as f32 - self.red as f32) * t) as u8,
+            green: (self.green as f32 + (other.green as f32 - self.green as f32) * t) as u8,
+            blue: (self.blue as f32 + (other.blue as f32 - self.blue as f32) * t) as u8,
         }
     }
 }
@@ -231,6 +308,8 @@ pub struct Style {
     fg_bright: bool,
     bg_bright: bool,
     attrs: Attributes,
+    fg_gradient: Vec<GradientColor>,
+    bg_gradient: Vec<GradientColor>,
     force: Option<bool>,
     for_stderr: bool,
 }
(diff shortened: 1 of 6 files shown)


djc:
This is a pretty large wad of new code, so I'd like some context on how you arrived at this being a good design that fits in with the current design of the library/API -- I understand the high-level goal of being accessible from a `dotted_string` and that this results in a `StyledObject` API. What does `on_` vs not `on_` mean?

src/utils.rs
```diff
@@ -114,6 +114,45 @@ impl Color {
     }
 }
 
+/// A Color used in a gradient
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct ColorGradient {
```
djc, line 119: Not a fan of this name, as it implies that `Gradient` is the noun here, when it's more like a `GradientColor`? Why can't you use the existing `Color` type here? How does support for gradients work on consoles that don't support as many colors?

src/utils.rs
```diff
@@ -114,6 +114,45 @@ impl Color {
...
+            blue: blue.min(255),
+        }
+    }
+
+    fn from_hex(hex: &str) -> Self {
+        // Remove '#' if present
+        let s = hex.trim_start_matches('#');
+
+        // 6 char hex
+        let channels: Vec<u8> = (0..5)
```
djc, line 139: Can we collect into a `[u8; 3]` instead of into a `Vec` directly?

src/utils.rs
```diff
@@ -114,6 +114,45 @@ impl Color {
...
+            .step_by(2)
+            .map(|i| u8::from_str_radix(&s[i..i + 2], 16).unwrap_or_default())
+            .collect();
+
+        ColorGradient::new(channels[0], channels[1], channels[2])
+    }
+
+    pub(crate) fn interpolate(&self, other: &ColorGradient, t: f32) -> ColorGradient {
```
djc, line 147: What is `t`? Use `Self` instead of repeating the type name.

src/utils.rs
```diff
@@ -618,7 +702,7 @@ impl<D> StyledObject<D> {
 }
 
 macro_rules! impl_fmt {
-    ($name:ident) => {
+    ($name:ident,$format_char:expr) => {
```
djc, line 705: Why is this necessary? Could it be extracted into a separate commit/PR?

src/utils.rs
```diff
@@ -630,32 +714,49 @@ macro_rules! impl_fmt {
...
-                        reset = true;
                     }
                     for attr in &self.style.attrs {
                         write!(f, "\x1b[{}m", attr.ansi_num())?;
                         reset = true;
                     }
                 }
-                fmt::$name::fmt(&self.val, f)?;
+                // Get the underlying value
+                let mut buf: String = format!($format_char, &self.val);
```
djc, line 747: Please avoid unnecessary type annotations.

src/utils.rs
```diff
@@ -732,6 +833,112 @@ fn char_width(c: char) -> usize {
     }
 }
 
+/// Returns the number of visible characters, ignoring style related chars
+fn count_visible_chars(text: &str) -> usize {
```
djc, line 837: I'm pretty sure we already have API for this (`measure_text_width()`), and this fails to consider various Unicode things.

src/utils.rs
```diff
@@ -732,6 +833,112 @@ fn char_width(c: char) -> usize {
...
+        }
+    }
+
+    total_visible
+}
+
+/// Applies the gradient over a string
+fn apply_gradient_impl(
+    text: &str,
+    gradients: &Vec<ColorGradient>,
```
djc, line 861: Should take a `&[_]` instead of `&Vec<_>`.

Mathias-Boulay:
*Note*: I'll break down the commit into logical units once the PR looks good. Otherwise, it is wasted time.

src/utils.rs
Mathias-Boulay: > Not a fan of this name, as it implies that Gradient is the noun here, when it's more like a GradientColor?

Indeed, 3 AM me was not the wisest on the naming.

> Why can't you use the existing Color type here? 

Adding another entry to the `Color` enum would not be backwards compatible with `match` statement without a default branch.

>How does support for gradients work on consoles that don't support as many colors?

At the time of your comment, badly. As of [bac82c](https://github.com/console-rs/console/commit/bac82cba0f537297eeb5a8b6ea8ef29aab150f08), it will simply not use the gradient colors.

src/utils.rs
Mathias-Boulay: Did a little something to use a fixed size array

src/utils.rs
Mathias-Boulay: It is indeed the case, it has been corrected.

src/utils.rs
Mathias-Boulay: Indeed, corrected

Mathias-Boulay:
> What does `on_` vs not `on_` mean?

It copies the existing API, where `on_` means the background is colored, else the text is. 
See functions like `black` and `on_black`, or the `from_dotted_string `allowing `[on_]<u8>`  for an arbitrary color for terminals supporting only 256 colors.

src/utils.rs
Mathias-Boulay: Well, that's probably the part I like the least about my PR, but I didn't manage otherwise.

Basically, to apply a gradient on a given string, I need to apply a color after each character.
So I needed to get the properly formatted `StyledObject<D>.value`.

Before the PR, the object was being directly printed into the formatter buffer via `fmt::$name::fmt(&self.val, f)?;`.

Now I get back a `String` from `format!(_)` macro, which require a way to format that data as an argument to delegate printing format to the same implementation (Display, Debug...).

If there is a way to get back the formatted string without such a hack, I would prefer it.

Mathias-Boulay:
Changes done and comments answered, now waiting for your input @djc :smile:

Mathias-Boulay:
When writing color detection for windows, it appeared that CMD/powershell were not respecting the TERM/COLORTERM variables. Other windows terms may, though.

src/utils.rs
```diff
@@ -114,6 +152,45 @@ impl Color {
...
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct GradientColor {
+    red: u8,
+    green: u8,
+    blue: u8,
+}
+
+impl GradientColor {
+    fn new(red: u8, green: u8, blue: u8) -> Self {
+        GradientColor {
```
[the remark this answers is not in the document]
Mathias-Boulay: Oh my, I forgot that u8 was limited to 255, correcting

src/utils.rs
```diff
@@ -22,10 +22,18 @@ fn default_colors_enabled(out: &Term) -> bool {
         || &env::var("CLICOLOR_FORCE").unwrap_or_else(|_| "0".into()) != "0"
 }
 
+fn default_true_colors_enabled(out: &Term) -> bool {
+    out.features().true_colors_supported()
+}
+
 static STDOUT_COLORS: Lazy<AtomicBool> =
     Lazy::new(|| AtomicBool::new(default_colors_enabled(&Term::stdout())));
+static STDOUT_TRUE_COLORS: Lazy<AtomicBool> =
```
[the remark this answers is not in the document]
Mathias-Boulay: It would work and allow to easily support more terminal color ranges. 
Unlikely to happen, as I've seen only one terminal with like 512. Probably pointless at this stage.

src/utils.rs
```diff
@@ -114,6 +152,45 @@ impl Color {
...
+impl GradientColor {
+    fn new(red: u8, green: u8, blue: u8) -> Self {
+        GradientColor {
+            red: red.min(255),
+            green: green.min(255),
+            blue: blue.min(255),
+        }
+    }
+
+    fn from_hex(hex: &str) -> Self {
```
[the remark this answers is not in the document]
Mathias-Boulay: I guess it doesn't kill to reduce template size. Implemented.

src/utils.rs
```diff
@@ -114,6 +152,45 @@ impl Color {
...
+        // Remove '#' if present
+        let s = hex.trim_start_matches('#');
+
+        // 6 char hex
+        let mut channels: [u8; 3] = Default::default();
+        for i in (0..5).step_by(2) {
+            channels[i / 2] = u8::from_str_radix(&s[i..i + 2], 16).unwrap_or_default()
+        }
+
+        GradientColor::new(channels[0], channels[1], channels[2])
```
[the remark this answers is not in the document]
Mathias-Boulay: Done

src/utils.rs
```diff
@@ -114,6 +152,45 @@ impl Color {
...
+            blue: blue.min(255),
+        }
+    }
+
+    fn from_hex(hex: &str) -> Self {
+        // Remove '#' if present
+        let s = hex.trim_start_matches('#');
+
+        // 6 char hex
+        let mut channels: [u8; 3] = Default::default();
```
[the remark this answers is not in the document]
Mathias-Boulay: `chunks_exact` isn't defined for `&str`, only `&[T]`. Did something nonetheless.

src/utils.rs
```diff
@@ -630,32 +758,50 @@ macro_rules! impl_fmt {
...
-                        reset = true;
                     }
                     for attr in &self.style.attrs {
                         write!(f, "\x1b[{}m", attr.ansi_num())?;
                         reset = true;
                     }
                 }
-                fmt::$name::fmt(&self.val, f)?;
+                // Get the underlying value
+                let mut buf = format!($format_char, &self.val);
```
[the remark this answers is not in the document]
Mathias-Boulay: Sadly there are extra allocations. This is related to [this comment about the additional macro argument](https://github.com/console-rs/console/pull/233#discussion_r1927111730). 
As good measure, I tweaked the code to have the allocation cost only for gradients, as it isn't needed for normal colors.

Mathias-Boulay:
Code updated, comments answered once again.

Waiting for more input on the new code/comments @djc :smile: 

unfinished comments conversation:
https://github.com/console-rs/console/pull/233#discussion_r1926908206
https://github.com/console-rs/console/pull/233#discussion_r1927111730
https://github.com/console-rs/console/pull/233#discussion_r1929225154
https://github.com/console-rs/console/pull/233#discussion_r1929230141

*Oh my god I sound like an AI with such formatting.*

Mathias-Boulay:
Heads up @djc you seem busy, is there someone else who can finish the review of the PR ?

Mathias-Boulay:
Hey there, any updates ?

djc:
Slowly catching up with my backlog.

Mathias-Boulay:
Ah, I see some conflicts appeared. 
Regardless of the conflicts, some additional constraints are being imposed by the const functions that removes my ability to use vectors. 

Should I dust off the PR, using a fixed size array instead ?

Mathias-Boulay:
I tried to wrangle with the const constraint but the code always ends up in a mess. I'll simply have my own version of console for the purpose of my application.

PR #234, opened 2025-02-20
fix: 🐛 remove double-width characters correctly

bestgopher:


djc:
Can you add a test?

bestgopher:
> Thanks, this looks good. Can you add a test?

Are there any examples for reference?

src/lib.rs
```diff
@@ -83,7 +83,7 @@ pub use crate::term::{
     user_attended, user_attended_stderr, Term, TermFamily, TermFeatures, TermTarget,
 };
 pub use crate::utils::{
-    colors_enabled, colors_enabled_stderr, measure_text_width, pad_str, pad_str_with,
+    char_width, colors_enabled, colors_enabled_stderr, measure_text_width, pad_str, pad_str_with,
```
djc, line 86: I'd prefer not to expose this.

src/utils.rs
```diff
@@ -719,7 +719,7 @@ fn str_width(s: &str) -> usize {
 }
 
 #[cfg(feature = "ansi-parsing")]
-fn char_width(c: char) -> usize {
+pub fn char_width(c: char) -> usize {
```
djc, line 722: Let's limit this to `pub(crate)` for now.

djc:
> > Thanks, this looks good. Can you add a test?
> 
> Are there any examples for reference?

Good question -- I guess we don't really do tests in this crate. Ignore my question!

src/lib.rs
bestgopher: Because I want to use it in https://github.com/console-rs/dialoguer/pull/321

src/utils.rs
djc, line 722: If we're going to make this `pub`, please fold the `not(feature = "ansi-parsing")` variant into this function (which I think happens implicitly anyway?) and add a docstring.

djc:
(If you rebase on main, CI should pass.)

src/utils.rs
bestgopher: Sorry, I didn't consider it carefully. I have now made this function pub(crate).

djc: So you don't need it in dialoguer anymore?

PR #235, opened 2025-02-26
Explicitly handle CtrlC on Windows

ericmarkmartin:
Make the Windows implementation of `read_single_key` respect the `ctrl_c_key` argument by un-setting the `ENABLE_PROCESS_INPUT` flag on the console mode.

We use RAII here to make a best-effort at resetting the mode to its previous state when done so as not to interfere with other applications that might want Ctrl-C to be processed by windows.

Tested by modifying the `keyboard` example to accept a flag `-r` | `--raw` which tells the example to use `read_key_raw` instead of `read_key` for consuming input. When running the example with this arg passed, we can hit Ctrl-C and see `CtrlC` printed out.

### DIFF
--- a/examples/keyboard.rs
+++ b/examples/keyboard.rs
@@ -3,10 +3,15 @@ use std::io;
 use console::{Key, Term};
 
 fn main() -> io::Result<()> {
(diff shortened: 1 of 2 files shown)


djc:
Thanks!

PR #244, opened 2025-03-05
Use EnumSet instead of a full-blown btreemap for the attributes

jwiesler:
This minimal change reduces the size of `Style` from 32 bytes to 10 and removes allocations needed to construct a style. This also means the remainder of the API could become `const`. If you want me to do that please ping me and I will do it in this mr.

The semantics of a bitset differ from the btreemap in so far as the btreemap is in _insertion order_, but I think this does not matter for attribute codes being printed, however, please verify this. Maybe this can be fixed by reordering the enum members.

The `EnumSetType` derive automatically derives `Copy`, `Clone`, `PartialEq` and `Eq` which is why I had to remove them.

For the library used see [enumset](https://crates.io/crates/enumset).

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -343,7 +343,7 @@ impl Style {
 
     /// Adds a attr.
     #[inline]
-    pub fn attr(mut self, attr: Attribute) -> Self {
+    pub const fn attr(mut self, attr: Attribute) -> Self {
         self.attrs = self.attrs.insert(attr);
         self
     }
@@ -435,39 +435,39 @@ impl Style {
     }
 
     #[inline]
-    pub fn bold(self) -> Self {
+    pub const fn bold(self) -> Self {
         self.attr(Attribute::Bold)
     }
     #[inline]
-    pub fn dim(self) -> Self {
+    pub const fn dim(self) -> Self {
         self.attr(Attribute::Dim)
     }
     #[inline]
-    pub fn italic(self) -> Self {
+    pub const fn italic(self) -> Self {
         self.attr(Attribute::Italic)
     }
     #[inline]
-    pub fn underlined(self) -> Self {
+    pub const fn underlined(self) -> Self {
         self.attr(Attribute::Underlined)
     }
     #[inline]
-    pub fn blink(self) -> Self {
+    pub const fn blink(self) -> Self {
         self.attr(Attribute::Blink)
     }
     #[inline]
-    pub fn blink_fast(self) -> Self {
+    pub const fn blink_fast(self) -> Self {
         self.attr(Attribute::BlinkFast)
     }
     #[inline]
-    pub fn reverse(self) -> Self {
+    pub const fn reverse(self) -> Self {
         self.attr(Attribute::Reverse)
     }
     #[inline]
-    pub fn hidden(self) -> Self {
+    pub const fn hidden(self) -> Self {
         self.attr(Attribute::Hidden)
     }
     #[inline]
-    pub fn strikethrough(self) -> Self {
+    pub const fn strikethrough(self) -> Self {
         self.attr(Attribute::StrikeThrough)
     }
 }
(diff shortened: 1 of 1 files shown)


djc:
> This minimal change reduces the size of `Style` from 32 bytes to 10 and removes allocations needed to construct a style. This also means the remainder of the API could become `const`.

Which part of this matters to you (most), what is your use case/motivation for this PR?

IMO enumset is pretty large dependency for something low-level like console. In particular it pulls in all the proc-macro machinery which I would prefer to avoid.

Maybe a more minimal "manual" implementation of the same principles could make sense.

jwiesler:
Mainly: Const API and no allocations/performance improvements.

I could do a manual implementation with u16 (maybe 30-60 LOC), but that would be harder for you to review :) Please tell me before I invest the time if you'd be willing to even merge this afterwards or under which constraints.

djc:
Happy to review 30-60 LoC of manual implementation, if it doesn't contain `unsafe` and doesn't bring in new dependencies.

(Maybe also look at the anstyle family of crates, including anstyle-stream.)

jwiesler:
I looked at anstyle-stream and implemented it in 60 LOC (barely). There are more changes, I added two tests and made all functions I found `const`.

src/utils.rs
```diff
@@ -116,35 +116,87 @@ impl Color {
...
+    Underlined = 3,
+    Blink = 4,
+    BlinkFast = 5,
+    Reverse = 6,
+    Hidden = 7,
+    StrikeThrough = 8,
 }
 
-impl Attribute {
+const ATTRIBUTES_LOOKUP: [Attribute; 9] = [
```
djc, line 134: Nit: let's make this `impl Attribute { const ALL: .. }` instead.

src/utils.rs
```diff
@@ -116,35 +116,87 @@ impl Color {
...
+        self.0 |= 1 << bit;
+    }
+
+    #[inline]
+    const fn bits(self) -> BitsIter {
+        BitsIter(self.0)
+    }
+
+    #[inline]
+    fn ansi_nums(self) -> impl Iterator<Item = u16> {
```
djc, line 167: I'm inclined to leave this out here and inline it into the one (library) callsite. What do you think?

src/utils.rs
```diff
@@ -116,35 +116,87 @@ impl Color {
...
+    }
+}
+
+struct BitsIter(u16);
+
+impl Iterator for BitsIter {
+    type Item = u16;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.0 == 0 {
```
djc, line 184: Nit, to reduce rightward drift, do an early return for the `0` case.

src/utils.rs
```diff
@@ -290,7 +342,7 @@ impl Style {
 
     /// Adds a attr.
     #[inline]
-    pub fn attr(mut self, attr: Attribute) -> Self {
+    pub const fn attr(mut self, attr: Attribute) -> Self {
```
djc, line 345: Nit: please extract all the `const` qualifiers into a separate commit, since they're unrelated.

djc (approved):
Looks great, just a few small nits!

jwiesler:
Phew: `mutable references are not allowed in constant functions`. That rust version 1.66 do be pretty old.

djc:
> Phew: `mutable references are not allowed in constant functions`. That rust version 1.66 do be pretty old.

Can probably just take `self` like the other methods?

jwiesler:
True, I thought those were the `attr()` functions.

src/utils.rs
jwiesler: I like it like this more, since `bits` would be a very opaque term on the calling site. But your call.

src/utils.rs
jwiesler: I prefer it like this, but I changed it.

PR #247, opened 2025-03-11
fix: clippy lint (clippy::unnecessary cast)

remi-dupre:
From https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_cast:

> warning: casting to the same type is unnecessary (`u16` -> `u16`)


djc:
Can you rebase this on main so the MSRV jobs will pass?

djc (approved):
Thanks!

PR #250, opened 2025-03-15
fix(utils): undefined behavior in truncate_str when tail is larger than width

remi-dupre:
Hello!

While looking again at #206 I bumped into a few issues in current implementation for `truncate_str`.

1. It may truncate if not required if `measure_width(s) + measure_width(tail) > width` : `truncate_str("foo bar", 7, "!") == "foo ba!"`. I don't think this is the expected behavior.
2. Using a tail longer than text width leads to overflowing substractions, which gives "undefined behavior" (ish) :
```rust
// In debug mode, panics:
truncate_str("foo bar", 2, "!!!")  # panics
// In release mode, weird result:
truncate_str("foo bar", 2, "!!!") == "foo bar"
```


djc (approved):
Makes sense, thanks!

PR #256, opened 2025-06-05
Add features to work with no_std, and with alloc in no_std

### LINKED ISSUE #255: Add no_std support
Some things, like `AnsiCodeIterator`, and `console::strip_ansi_codes` should work with `no_std`.

ChocolateLoverRaj:
Closes #255 

3 levels:
no_std without alloc: some stuff works
no_std with alloc: some more stuff works
std: everything works

Tested by using `console::strip_ansi_codes` in [my OS](https://github.com/ChocolateLoverRaj/rust-os-tutorial), on `x86_64-unknown-none` with the `alloc` feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,13 +13,15 @@ readme = "README.md"
 rust-version = "1.66"
 
 [features]
-default = ["unicode-width", "ansi-parsing"]
+std = ["dep:libc", "dep:once_cell", "alloc"]
+alloc = []
+default = ["unicode-width", "ansi-parsing", "std"]
 windows-console-colors = ["ansi-parsing"]
 ansi-parsing = []
 
 [dependencies]
-libc = "0.2.99"
-once_cell = "1.8"
+libc = { version = "0.2.99", optional = true }
+once_cell = { version = "1.8", optional = true }
 unicode-width = { version = "0.2", optional = true }
 
 [target.'cfg(windows)'.dependencies]
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -1,5 +1,6 @@
-use std::{
-    borrow::Cow,
+#[cfg(feature = "alloc")]
+use alloc::{borrow::Cow, string::String};
+use core::{
     iter::{FusedIterator, Peekable},
     str::CharIndices,
 };
@@ -185,6 +186,7 @@ fn find_ansi_code_exclusive(it: &mut Peekable<CharIndices>) -> Option<(usize, us
     }
 }
 
+#[cfg(feature = "alloc")]
 /// Helper function to strip ansi codes.
 pub fn strip_ansi_codes(s: &str) -> Cow<str> {
     let mut char_it = s.char_indices().peekable();
--- a/src/kb.rs
+++ b/src/kb.rs
@@ -1,3 +1,5 @@
+use alloc::vec::Vec;
+
 /// Key mapping
 ///
 /// This is an incomplete mapping of keys that are supported for reading
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,25 +77,37 @@
 //!   calculations).
 
 #![warn(unreachable_pub)]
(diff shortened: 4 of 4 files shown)


Cargo.toml
```diff
@@ -13,13 +13,15 @@ readme = "README.md"
 rust-version = "1.66"
 
 [features]
-default = ["unicode-width", "ansi-parsing"]
+std = ["dep:libc", "dep:once_cell", "alloc"]
```
djc, line 16: Nit: please keep `default` first.

Makefile
```diff
@@ -13,9 +13,9 @@ test:
 	@echo "CARGO TESTS"
 	@cargo test
 	@cargo test --all-features
-	@cargo test --no-default-features
-	@cargo test --no-default-features --features ansi-parsing
-	@cargo test --no-default-features --features unicode-width
+	@cargo test --no-default-features --features std
```
djc, line 16: Nit: please add a line that tests for `--no-default-features` without `std`, and also one with `--no-default-features --features alloc`.

src/ansi.rs
```diff
@@ -185,6 +186,7 @@ fn find_ansi_code_exclusive(it: &mut Peekable<CharIndices>) -> Option<(usize, us
     }
 }
 
+#[cfg(feature = "alloc")]
```
djc, line 189: Nit: move the `cfg` below the docstring.

djc (approved):
A few more nits -- should be straightforward. Please squash your changes into a single commit.

Makefile
ChocolateLoverRaj: If I run `cargo test --no-default-features`, I get errors like this:
```
error[E0432]: unresolved import `console::style`
  --> examples/colors.rs:1:5
   |
1  | use console::style;
   |     ^^^^^^^^^^^^^^ no `style` in the root
   |
note: found an item that was configured out
  --> /home/rajas/Documents/console/src/lib.rs:97:52
   |
97 |     set_colors_enabled, set_colors_enabled_stderr, style, truncate_str, Alignment, Attribute,
   |                                                    ^^^^^
note: the item is gated behind the `std` feature
  --> /home/rajas/Documents/console/src/lib.rs:94:7
   |
94 | #[cfg(feature = "std")]
   |       ^^^^^^^^^^^^^^^
```
So then I added:
```
[[example]]
name = "colors"
required-features = ["std"]

[[example]]
name = "colors256"
required-features = ["std"]

[[example]]
name = "cursor_at"
required-features = ["std"]

[[example]]
name = "keyboard"
required-features = ["std"]

[[example]]
name = "term"
required-features = ["std"]
```
But then I get errors from the comments at the top of `lib.rs`:
```
error[E0432]: unresolved import `console::style`
  --> src/lib.rs:44:5
   |
5  | use console::style;
   |     ^^^^^^^^^^^^^^ no `style` in the root
   |
note: found an item that was configured out
  --> /home/rajas/Documents/console/src/lib.rs:97:52
   |
97 |     set_colors_enabled, set_colors_enabled_stderr, style, truncate_str, Alignment, Attribute,
   |                                                    ^^^^^
note: the item is gated behind the `std` feature
  --> /home/rajas/Documents/console/src/lib.rs:94:7
   |
94 | #[cfg(feature = "std")]
   |       ^^^^^^^^^^^^^^^
```
So I don't know what to do.

ChocolateLoverRaj:
I squashed and force pushed.

Makefile
ChocolateLoverRaj: > Maybe `cargo test --lib --no-default-features`?

This works, but 0 tests get run, so I don't see the point of this.

djc:
Looks like the WASM situation needs a little work. Otherwise this is looking good!

ChocolateLoverRaj:
I made wasm and windows require the "std" feature now so it should work.

djc (approved):

djc:
Thanks!

PR #258, opened 2025-06-05
Add WithoutAnsi struct that implements Display

### LINKED ISSUE #257: Add a struct that strips ANSI and implements `Display`
`console::strip_ansi_codes` copies the entire string if it contains any ANSI. In cases where the stripped string needs to be written to something that implements `Write`, copying the string is completely unnecessary. If there was a struct like `WithoutAnsi` which implemented `Display`, it could be used to efficiently write to a serial port (my use case) without copying data.

I can make a PR for this, just let me know how the API should be. Imo it should be something like
```rs
pub struct WithoutAnsi;
impl<'a> WithoutAnsi<'a> {
  pub fn new(&'a str) -> Self;
}
```

ChocolateLoverRaj:
Closes #257 

~~Making this a draft PR because #255 should be merged first.~~

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -1,6 +1,7 @@
 #[cfg(feature = "alloc")]
 use alloc::{borrow::Cow, string::String};
 use core::{
+    fmt::Display,
     iter::{FusedIterator, Peekable},
     str::CharIndices,
 };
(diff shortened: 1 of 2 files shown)


src/without_ansi.rs
```diff
@@ -0,0 +1,41 @@
+use core::fmt::Display;
+
+use crate::AnsiCodeIterator;
+
+/// A wrapper struct that implements [`core::fmt::Display`], only displaying non-ansi parts.
+pub struct WithoutAnsi<'a> {
```
djc, line 6: I don't want this in a separate module. Please move it just above the `AnsiCodeIterator` definition.

djc (approved):

PR #269, opened 2025-08-22
feat: add get cursor position

dhuseby:
This adds the ability to get the current cursor position using the ansi cursor
position query sequence. Sometimes it is useful to know where you are in the
terminal when your program first starts up, or when you are doing some complex
terminal manipulation.

Signed-off-by: Dave Grantham <dwg@linuxprogrammer.org>

### DIFF
--- a/examples/cursor_position.rs
+++ b/examples/cursor_position.rs
@@ -1,0 +1,50 @@
+extern crate console;
+use console::Term;
+use std::io;
+
+fn draw_point(term: &Term, width: usize) -> io::Result<()> {
+    // no cheating...get the cursor position here
+    let (x, y) = term.cursor_position()?;
+    let str = format!("({x}, {y})");
+    let w = str.len() + 2;
+    if x + w > width {
+        term.move_cursor_left(w - 1)?;
+        term.write_str(&format!("{str} •"))?;
+    } else {
+        term.write_str(&format!("• {str}"))?;
+    }
+    Ok(())
+}
+
+fn main() -> io::Result<()> {
+    let term = Term::stdout();
+    term.hide_cursor()?;
+    term.clear_screen()?;
+
+    let (height, width): (usize, usize) = (term.size().0 as usize, term.size().1 as usize);
+
+    // draw the four corners
+    term.move_cursor_to(0, 0)?;
+    draw_point(&term, width)?;
+    // this tests the formatting logic
+    for i in 0..20 {
+        term.move_cursor_to(width - i - 1, i)?;
+        draw_point(&term, width)?;
+    }
+    term.move_cursor_to(0, height - 2)?;
+    draw_point(&term, width)?;
+    term.move_cursor_to(width, height - 2)?;
+    draw_point(&term, width)?;
+
+    for _ in 0..10 {
+        let x = rand::random_range(..=width - 1);
+        let y = rand::random_range(1..=height - 3);
+        term.move_cursor_to(x, y)?;
+        draw_point(&term, width)?;
+    }
+
+    term.move_cursor_to(0, height)?;
+    term.show_cursor()?;
+
+    Ok(())
+}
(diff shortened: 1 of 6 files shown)


examples/cursor_position.rs
```diff
@@ -0,0 +1,50 @@
+extern crate console;
```
djc, line 1: Please drop the added example.

src/ansi.rs
```diff
@@ -189,7 +189,7 @@ fn find_ansi_code_exclusive(it: &mut Peekable<CharIndices>) -> Option<(usize, us
 
 /// Helper function to strip ansi codes.
 #[cfg(feature = "alloc")]
-pub fn strip_ansi_codes(s: &str) -> Cow<str> {
+pub fn strip_ansi_codes(s: &str) -> Cow<'_, str> {
```
djc, line 192: Since this is independent of your other changes, please isolate it in a separate commit.

src/unix_term.rs
```diff
@@ -335,6 +373,23 @@ fn read_single_key_impl(fd: RawFd) -> Result<Key, io::Error> {
     }
 }
 
+pub(crate) fn cursor_position() -> io::Result<(usize, usize)> {
+    // Send the cursor position request escape sequence
+    print!("\x1b[6n");
+    io::stdout().flush()?;
+
+    // Read the response from the terminal
+    let key = read_single_key(false)?;
```
djc, line 382: Nit: please inline the `key` binding.

djc (commented):
Here's some initial feedback. It doesn't look like this needs a semver-incompatible version bump. It does look like it needs either a Windows implementation or a Unix-only guard.

examples/cursor_position.rs
djc: I don't think think every tiny feature merits an example that then has to be maintained forever.

src/unix_term.rs
dhuseby: fixed

dhuseby:
reverted back to 0.16.0 as well.

src/unix_term.rs
```diff
@@ -258,6 +258,44 @@ fn read_single_key_impl(fd: RawFd) -> Result<Key, io::Error> {
                                 'H' => Ok(Key::Home),
                                 'F' => Ok(Key::End),
                                 'Z' => Ok(Key::BackTab),
+                                '0'..='9' => {
```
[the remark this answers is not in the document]
dhuseby: I could refactor this to reduce the indentation at several levels. I was trying to keep my edits minimal. If it were me I would refactor the block to be something like:

```
loop {
  match read_single_char(fd)? {
    Some('\x1b') => break match read_single_char(fd)? {
      Some('[') => match read_single_char(fd)? {
        Some(c) => match c {
          'A' => Ok(Key::ArrowUp),
          //...
          '0'..='9' => read_cursor_pos(c),
          _ => read_tilde_seq(c),
        }
        _ => Ok(Key::UnknownEscSeq(vec![c])),
      }
      Some(c) => Ok(Key::UnknownEscSeq(vec![c])),
      _ => Ok(Key::Escape)
    }
    Some(c) => read_non_escape_seq(c),
    None => {
      // there is no subsequent byte ready to be read, block and wait for
      // input negative timeout means that it will block indefinitely
      match select_or_poll_term_fd(fd, -1) {
        Ok(_) => continue,
        Err(_) => break Err(io::Error::last_os_error()),
      }
    }
  }
}

fn read_cursor_pos(c: char) -> Result<Key, io::Error> {
  //...
}

fn read_tilde_seq(c: char) -> Result<Key, io::Error> {
  //...
}

fn read_non_escape_seq(c: char) -> Result<Key, io::Error> {
  //...
}
```

Do you think moving the match arm bodies to functions is easier to read?

PR #284, opened 2026-03-30
chore: Remove redudant helper and add `TermInner` constructors

ChrisDenton:
`windows-sys` long ago moved to using a std compatible raw handle type.

Adding constructors to `TermInner` makes it much easier to experiment with its implementation.


djc (approved):
Thanks!

PR #286, opened 2026-05-05
fix(unix): retry EINTR in poll/select/read

### LINKED ISSUE #212: SIGWINCH generates SIGINT
If we look at the code located here https://github.com/console-rs/console/blob/master/src/unix_term.rs#L325
you should realize that `EINTR` can be generated also by other signals than `SIGINT`.
In my application, if I resize the terminal, a `SIGWINCH` signal is thrown, but your crate translates that `SIGWINCH` (which generates an `EINTR`) to a `SIGINT`, so you're throwing an unwanted `SIGINT` to the "outselves", causing my application to brutally exit (since it received a `SIGINT`).
PLEASE FIX THIS!

jh-block:
Fixes #212.

`read_single_key` treats any `Err(Interrupted)` from `read_single_key_impl` as Ctrl-C and forwards it via `libc::raise(SIGINT)`. That's correct only for the synthesized `Err(Interrupted)` `read_bytes` returns when it sees a `\x03` byte. EINTR returned by `poll`/`select`/`read` because of an unrelated signal (SIGCHLD, SIGWINCH, etc.) takes the same path and incorrectly kills the program with SIGINT.

Retry EINTR inside the syscall wrappers so `Err(Interrupted)` reaching `read_single_key` unambiguously means a `\x03` byte was read.

### DIFF
--- a/src/unix_term.rs
+++ b/src/unix_term.rs
@@ -157,21 +157,25 @@ fn poll_fd(fd: RawFd, timeout: i32) -> io::Result<bool> {
         events: libc::POLLIN,
         revents: 0,
     };
-    let ret = unsafe { libc::poll(&mut pollfd as *mut _, 1, timeout) };
-    if ret < 0 {
-        Err(io::Error::last_os_error())
-    } else {
-        Ok(pollfd.revents & libc::POLLIN != 0)
+    loop {
+        let ret = unsafe { libc::poll(&mut pollfd as *mut _, 1, timeout) };
+        if ret < 0 {
+            let err = io::Error::last_os_error();
+            if err.kind() == io::ErrorKind::Interrupted {
+                continue;
+            }
+            break Err(err);
+        }
+        break Ok(pollfd.revents & libc::POLLIN != 0);
     }
 }
 
 #[cfg(target_os = "macos")]
 fn select_fd(fd: RawFd, timeout: i32) -> io::Result<bool> {
-    unsafe {
-        let mut read_fd_set: libc::fd_set = mem::zeroed();
-
+    loop {
+        let mut read_fd_set: libc::fd_set = unsafe { mem::zeroed() };
         let mut timeout_val;
-        let timeout = if timeout < 0 {
+        let timeout_ptr: *mut libc::timeval = if timeout < 0 {
             ptr::null_mut()
         } else {
             timeout_val = libc::timeval {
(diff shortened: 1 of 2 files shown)


djc:
Thanks for the clippy fix -- can you split it out in a separate commit please?

For the first hunk's loop, I think we could do `break` instead of `return` at the end? Seems slightly nicer.

djc:
This does not seem like a clean split. Also, the first commit looks like it makes some unrelated/unnecessary changes. If you are using an LLM to help with this, please self-review before asking for my review.

jh-block:
Can you clarify what you mean? The only perhaps-unrelated change in the first commit is reducing the scope of the unsafe block in `select_fd` to the actual unsafe code, rather than covering the whole method with it, which seemed appropriate since otherwise the entire loop (which is just normal safe logic) would be within the unsafe block. Do you want me to revert that, or split it into a separate commit, or are you referring to something else?

jh-block:
@djc is there any chance you could take another look at the two commits? I am not sure how to interpret your earlier comment; the split of the clippy fix and the logic fix into two commits _is_ clean, and the first commit (the EINTR logic fix) contains only the changes I intended to make. If you would like something changed can you indicate what it is?

djc:
This hunk from the first commit seems like an unnecessary increase in scope of the `unsafe` block?

Screenshot 2026-05-09 at 14 10 27

src/unix_term.rs
```diff
@@ -157,21 +157,25 @@ fn poll_fd(fd: RawFd, timeout: i32) -> io::Result<bool> {
...
-    } else {
-        Ok(pollfd.revents & libc::POLLIN != 0)
+    loop {
+        let ret = unsafe { libc::poll(&mut pollfd as *mut _, 1, timeout) };
+        if ret < 0 {
+            let err = io::Error::last_os_error();
+            if err.kind() == io::ErrorKind::Interrupted {
+                continue;
+            }
+            break Err(err);
```
djc, line 167: Nit: suggest writing this as

```rust
match err.kind() {
    io::ErrorKind::Interrupted => continue,
    _ => break Err(err),
}
```

src/unix_term.rs
```diff
@@ -181,20 +185,25 @@ fn select_fd(fd: RawFd, timeout: i32) -> io::Result<bool> {
...
+        };
         if ret < 0 {
-            Err(io::Error::last_os_error())
-        } else {
-            Ok(libc::FD_ISSET(fd, &read_fd_set))
+            let err = io::Error::last_os_error();
+            if err.kind() == io::ErrorKind::Interrupted {
+                continue;
+            }
+            break Err(err);
```
djc, line 204: Same here.

PR #287, opened 2026-05-11
Style::from_dotted_str ignore non-ascii

danjl1100:
True colors were added to `Style::from_dotted_str` in [v0.16.2](https://github.com/console-rs/console/compare/0.16.1...0.16.2), but it seems like the string slicing only works for ascii (panicking for certain inputs).

Of course, only ascii is expected as an input to the function, but I figured I'd raise this PR since the [doc for that function](https://docs.rs/console/0.16.3/console/struct.Style.html#method.from_dotted_str) ends with "Unknown terms are ignored".  It seems unlikely (possible?) that someone would provide user-input to this function. At least the length check ensures we're not linearly checking ascii in any huge strings.

- The first commit demonstrates the panic in the current code with `#[expect_panic]` tests.
- The second commit checks for ascii in the offending branches and removes `#[expect_panic]` to verify  "Unknown terms are ignored".

---
I identified this improvement while reviewing a delta in cargo-vet (myself).

I used AI to create the input strings that trigger both panic branches (as it's more familiar with UTF-8 details than me), but the commits and asserts are my own.  I debated removing the comments the AI generated above the two inputs, but it seems like it helps document (at least against the current implementation).  Feel free to adjust as you see fit.


danjl1100:
Hm.. I'm not sure why the lint run on windows failed (the others passed)

djc:
> Hm.. I'm not sure why the lint run on windows failed (the others passed)

Pre-existing, should be fixed if you rebase:

- https://github.com/console-rs/console/pull/288

djc (approved):
Makes sense, thanks!

danjl1100:
Awesome, rebased and the checks look better now.

PR #290, opened 2026-06-30
feat: add `is_dumb` public function

ugur-eren:
# Add public `is_dumb()` function
Exposes a new public function `is_dumb()` that provides a fast way to check whether the application has a "dumb" terminal (i.e. `TERM` is unset or set to `dumb` on Unix; `TERM` explicitly set to `dumb` on Windows).

This logic previously existed only as an inline check inside `is_a_color_terminal()`. This PR extracts it into a dedicated `is_a_dumb_terminal()` function per platform (`unix_term.rs`, `windows_term/mod.rs`, `wasm_term.rs`), and wires it up through a new public `is_dumb()` in `term.rs`, re-exported from `lib.rs`.

## Behavior
No change to existing `is_a_color_terminal()` output. This is a pure refactor for that path. New behavior is only the addition of the standalone `is_dumb()` API.

## Note
`is_dumb()` semantics differ slightly by platform when `TERM` is unset (Unix treats unset as dumb but Windows does not, since native Windows terminals typically don't set `TERM`).

This is intentional and documented on the function.

This PR is in relation to conversation on PR console-rs/indicatif#808

### DIFF
--- a/src/term.rs
+++ b/src/term.rs
@@ -1,5 +1,6 @@
 use alloc::sync::Arc;
 use core::fmt::{Debug, Display};
+use std::env;
 use std::io::{self, Read, Write};
 #[cfg(any(unix, all(target_os = "wasi", target_env = "p1")))]
 use std::os::fd::{AsRawFd, RawFd};
@@ -565,6 +566,26 @@ impl Term {
             }
         }
         Ok(())
+    }
+}
+
+/// A fast way to check if the application has a dumb terminal.
+///
+/// On Unix: `TERM` environment variable is not set or set to `dumb`.
+///
+/// On Windows: `TERM` environment variable is explicitly set to `dumb`.
+/// Native windows terminals typically do not set the `TERM` environment variable.
+#[inline]
+pub fn is_dumb() -> bool {
+    #[cfg(windows)]
+    let default = false;
+
+    #[cfg(not(windows))]
+    let default = true;
+
+    match env::var("TERM") {
+        Ok(term) => term == "dumb",
+        Err(_) => default,
     }
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "console"
 description = "A terminal and console abstraction for Rust"
-version = "0.16.3"
+version = "0.16.4"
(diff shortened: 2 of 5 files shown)


src/term.rs
```diff
@@ -568,6 +568,17 @@ impl Term {
...
 
+/// A fast way to check if the application has a dumb terminal.
+///
+/// On Unix: `TERM` environment variable is not set or set to `dumb`.
+///
+/// On Windows: `TERM` environment variable is explicitly set to `dumb`.
+/// Native windows terminals typically do not set the `TERM` environment variable.
+#[inline]
+pub fn is_dumb() -> bool {
+    is_a_dumb_terminal()
```
djc: Suggest to inline the `is_a_dumb_terminal()` logic here, with a platform-dependent input `#[cfg(windows)] let default = true;` value that is returned in the `Err` case.

src/wasm_term.rs
```diff
@@ -21,6 +21,11 @@ pub(crate) fn is_a_terminal(_out: &Term) -> bool {
     }
 }
 
+#[inline]
+pub(crate) fn is_a_dumb_terminal() -> bool {
+    false
```
djc: Why is `false` the right default for WASM?

ugur-eren: You're right, sorry. Since it doesn't support ANSI escape sequences, it should return `true`. With the last commit, it will return `true`

djc:
Can you squash your existing commits and add a version bump commit?

djc (approved):

ugur-eren:
> add a version bump commit

Should I increase patch or minor? Sorry, I don't know how cargo treats pre-1.0 releases.

djc:
> > add a version bump commit
> 
> Should I increase patch or minor? Sorry, I don't know how cargo treats pre-1.0 releases.

No worries. It should be 0.16.4, since this PR does not make any semver-incompatible changes. If this crate was post-1.0.0 we'd increase the minor version because the new API is a forward compatibility hazard.

djc:
Published [0.16.4](https://github.com/console-rs/console/releases/tag/0.16.4), thanks!

PR #292, opened 2026-07-15
Add OSC 9;4 (term/OS progress indicator) support

epage:
This is to unblock console-rs/indicatif#596

The API is modeled off of color support with the one exception of not providing a`Term::term_progress`. 

There is no standard name for OSC 9;4. I wasn't the most consistent in naming
- "progress integration" (for a terminal)
- `TermProgress`

In Cargo, we called the config field [`term.progress.term-integration`](https://doc.rust-lang.org/stable/cargo/reference/config.html#termprogressterm-integration).

This delegates the terminal detection logic and ANSI escape code writing to `anstyle-progress` which is a purpose-built crate and should not expand in scope to pull in other dependencies (unless the crate is being split even smaller). There are no de facto standard override variables for term progress like no-color.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,12 +14,13 @@ include = ["CHANGELOG.md", "Cargo.toml", "LICENSE", "README.md", "src/**/*.rs"]
 
 [features]
 default = ["unicode-width", "ansi-parsing", "std"]
-std = ["dep:libc", "alloc"]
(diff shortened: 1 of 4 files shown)


epage:
Starting as a draft so I can collect any feedback people want to give while I work on the indicatif side of this

djc (approved):
This looks okay to me.

PR #294, opened 2026-08-13
Honor NO_COLOR without suppressing text attributes

### LINKED ISSUE #291: NO_COLOR drops non-color attributes and treats empty value as active
# `NO_COLOR` drops non-color attributes and treats empty value as active

## Summary

`console` honors `NO_COLOR` but in two ways that diverge from
[no-color.org](https://no-color.org/)
(or [its source](https://github.com/jcs/no_color/blob/master/index.md)):

1. It suppresses **all** styling — bold, italic, underline, dim, reverse,
   blink, hidden, strikethrough — not only color, contrary to the FAQ:

   > **Q:** Should the presence of `NO_COLOR` disable other styling such as
   > bold, underline, and italic?
   >
   > **A:** No. This standard only signals the user's intention regarding
   > adding ANSI color to text output.

2. It treats `NO_COLOR=""` (present but empty) as active. The spec requires
   present **and not an empty string**; an empty value disables nothing.

## Reproduction

Both bugs are in `StyledObject::fmt` (`src/utils.rs`, the `impl_fmt!` macro):
the entire emit block — fg, bg, **and** `self.style.attrs` — sits behind a
single `colors_enabled()` / `colors_enabled_stderr()` gate, which flips to
`false` when `env::var("NO_COLOR").is_ok()` (`src/unix_term.rs`,
`src/windows_term/mod.rs`).

Verified on `console` 0.16.4 — `style("X").fg(Color::Red).attr(Attribute::Bold)`:

```rust
set_colors_enabled(true);  let on = format!("{st}{s}{st:#}");  // "\x1b[31m\x1b[1mX..."
set_colors_enabled(false); let off = format!("{st}{s}{st:#}"); // "X" — bold gone too
```

`off` contains neither `\x1b[31m` (red) nor `\x1b[1m` (bold). Per the FAQ,
only the color should be gone; bold should survive.

## Expected (spec-aligned)

- **#1 scope:** When `NO_COLOR` is active, suppress color only; keep emitting
  `self.style.attrs` (bold/dim/italic/underline/blink/reverse/hidden/strikethrough).
  `Style` already separates `fg`/`bg` from `attrs`, so the fix is to move the
  `attrs` emit outside the `colors_enabled()` gate.
- **#2 trigger:** `NO_COLOR=""` should disable nothing. Replace
  `env::var("NO_COLOR").is_ok()` with a non-empty check
  (e.g. `env::var_os("NO_COLOR").is_some_and(|v| !v.is_empty())`).

## Notes

- `crossterm` is a working example of color-only `NO_COLOR` handling — it
  gates fg/bg/underline-color but emits `SetAttributes` unconditionally.
- This cascades to dependents using `console` for styling — `dialoguer` and
  `indicatif` — which would inherit the fix.
- I'm framing this as a bug fix (patch bump), but defer to the project's
  own version policy.

MsfPablo:
## Summary

The current `StyledObject` formatting path in `console` drops the
attribute escapes (bold, italic, underline, blink, reverse, hidden,
strikethrough) whenever colors are disabled — and `NO_COLOR` (per
no-color.org) is what disables colors. That contradicts the spec, which
is explicit that only color is suppressed:

> Command-line software which adds ANSI color to its output by default
> should check for a NO_COLOR environment variable that, when present
> (regardless of its value), prevents the addition of ANSI color.

`console` also currently disables color when `NO_COLOR=""` is set,
because `env::var("NO_COLOR").is_ok()` returns `true` for any value
(including the empty string). Per the spec the presence of the variable
alone should disable color — but a deliberately-empty variable should be
a no-op.

## Changes

* Move the attribute escape emission in `impl_fmt!` (src/utils.rs)
  outside the `colors_enabled()` / `colors_enabled_stderr()` gate, so
  attributes are still emitted (with a trailing reset) when colors are
  suppressed.
* Replace `env::var("NO_COLOR").is_ok()` with a non-empty check
  (`env::var_os("NO_COLOR").is_some_and(|v| !v.is_empty())`) in
  `src/unix_term.rs` and `src/windows_term/mod.rs`, matching the
  spec.
* Add a regression test (`test_attrs_survive_colors_disabled`) covering
  both behaviors.

## Verification

```sh
$ make test          # 12 test runs, all pass
$ make lint          # clippy with --deny warnings, clean
$ make format-check  # rustfmt --check, clean
```

The new test verifies:

* `style("foo").bold()` still emits `\x1b[1m…\x1b[0m` when
  `colors_enabled()` is `false`.
* `style("bar").italic()` still emits `\x1b[3m…\x1b[0m` in the
  same condition.
* `style("baz").red()` does **not** emit `\x1b[31m` and does
  **not** emit a trailing reset when colors are disabled.

## AI disclosure

This PR was drafted with AI assistance (Claude) and self-reviewed. I
read the relevant code paths, reproduced the failing behavior locally
against the crate's existing test harness, and verified the fix does not
change the rendered output for the unaffected code paths.

Fixes #291


src/windows_term/mod.rs
```diff
@@ -63,7 +63,7 @@ pub(crate) fn is_a_color_terminal(out: &Term) -> bool {
     if !is_a_terminal(out) {
         return false;
     }
-    if env::var("NO_COLOR").is_ok() {
+    if env::var_os("NO_COLOR").is_some_and(|v| !v.is_empty()) {
```
djc, line 66: I would like these changes to be in a separate commit. Should we also check for `v.trim() == '0'`?

djc (approved):

PR #295, opened 2026-08-24
Add a Link type for printing OSC 8 hyperlinks

### LINKED ISSUE #132: Print hyperlinks
https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

The list of applications supporting this is increasing, even the Windows terminal appears to support now.

ChrisJr404:
Closes #132.

console already recognizes OSC 8 hyperlinks when stripping/measuring text, but there was no way to emit one. This adds a small `Link` wrapper, in the same spirit as `Emoji`, that prints text as an OSC 8 hyperlink on terminals that support it.

Like styling, the escape is only written when colors are enabled for the target stream, so piped or redirected output just shows the plain text. You can force it on or off with `.force(..)` and target stderr with `.for_stderr()`.

```rust
use console::Link;
println!("Read the {}.", Link::new("docs", "https://docs.rs/console"));
```

I went with `Link::new(text, url)` so it reads a bit like a Markdown link, but I'm happy to flip the order if you'd rather have the url first. epage linked anstyle-hyperlink on the issue as a reference; I kept this minimal for now and left out the optional `id=` parameter.


src/utils.rs
```diff
@@ -875,6 +875,87 @@ impl fmt::Display for Emoji<'_, '_> {
...
+    ///
+    /// This overrides the automatic detection that otherwise follows the color
+    /// configuration of the target stream.
+    pub fn force(mut self, value: bool) -> Link<'a, D> {
+        self.force = Some(value);
+        self
+    }
+
+    fn enabled(&self) -> bool {
+        self.force.unwrap_or_else(|| match self.for_stderr {
```
djc, line 942: Since this isn't really about color, I don't think we should disable the links for this.

src/utils.rs
```diff
@@ -875,6 +875,87 @@ impl fmt::Display for Emoji<'_, '_> {
...
+    }
+}
+
+impl<D: fmt::Display> fmt::Display for Link<'_, D> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.enabled() {
+            write!(f, "\x1b]8;;{}\x1b\\{}\x1b]8;;\x1b\\", self.url, self.text)
+        } else {
+            self.text.fmt(f)
+        }
```
djc, line 955: Nit: suggest writing this as

```rust
match self.enabled() {
    true => write!(..),
    false => self.text.fmt(f),
}
```

PR #296, opened 2026-09-01
Fix truncate_str panicking mid-character without ansi-parsing

lenamonj:
`truncate_str` panics on multi-byte input when built without `ansi-parsing`:

```rust
// default-features = false, features = ["std", "unicode-width"]
console::truncate_str("你好世界", 5, "");
// panicked at src/utils.rs:978:15:
// end byte index 5 is not a char boundary; it is inside '好'
```

The `not(ansi-parsing)` branch slices with `&s[..width.saturating_sub(tail.len())]`: a byte offset used as a column count, and the tail charged by `len()` instead of its display width. It now walks chars against a budget of `width - str_width(tail)` and cuts on a boundary, matching the `ansi-parsing` branch.

That also merges `char_width`'s two definitions into one gated on `unicode-width`. The `not(ansi-parsing)` copy returned 1 for every char, so it disagreed with `str_width` in builds that enable `unicode-width`.

Tests are gated per feature state like the ones beside them; both fail on main with the panic above. All seven `make test` configurations pass, plus `cargo fmt --check` and `cargo clippy`.

Two related bugs in the same class are not in this PR: `read_line_initial_text` mixes `initial.len()` with a char index (non-ASCII initial text discards typed input), and Windows `read_secure` backspace uses `rv.truncate(rv.len() - 1)`. Happy to file either separately.

Found with AI assistance; verified and reviewed by me.


src/utils.rs
```diff
@@ -973,11 +972,24 @@ pub fn truncate_str<'a>(s: &'a str, width: usize, tail: &str) -> Cow<'a, str> {
...
-            "{}{}",
-            &s[..width.saturating_sub(tail.len())],
-            tail
-        ))
+        // Walk by display width and stop on a char boundary: indexing by
+        // `width` treats a byte offset as a column count and panics mid-char.
+        let budget = width.saturating_sub(str_width(tail));
+        let mut end = 0;
+        let mut used = 0;
+        for c in s.chars() {
```
djc: I think we can leverage [`char_indices()`](https://doc.rust-lang.org/std/primitive.str.html#method.char_indices) here?

src/utils.rs
```diff
@@ -973,11 +972,24 @@ pub fn truncate_str<'a>(s: &'a str, width: usize, tail: &str) -> Cow<'a, str> {
...
+            let cw = char_width(c);
+            if used + cw > budget {
```
djc: I think this could use some comments on the subtleties of the difference between `char_width()` and `len_utf8()`, and I don't think `cw` and `used` and `end` are the best names for bindings to represent these subtleties.

src/utils.rs
lenamonj: Done, uses `char_indices()` now.

src/utils.rs
lenamonj: Renamed to `column_budget`, `columns`, `cut_at`, `byte_index` and added a comment on columns vs byte offsets. All seven `make test` configurations pass.

djc (approved):
Much better, thanks!

PR #297, opened 2026-09-05
perf: accelerate printable ASCII text width

dexhunter:
`measure_text_width` currently runs printable ASCII through ANSI parsing and width calculation. This adds a boolean byte reduction that returns the byte length for printable ASCII when ANSI parsing is enabled. Other inputs retain the existing parser and width behavior.

On Rust 1.97.1 / Linux x86_64, a fixed five-corpus public-API microbenchmark improved from 320.91 ns to 90.79 ns (71.7% lower); a separate winner replay measured 90.58 ns. The mix includes three ASCII corpora, Unicode status text and colored status text. The latter two cases were about 7.0% and 2.8% slower, respectively; all eight fallback/tiny regression guards passed. These are function-latency measurements.

Validation passed all seven native feature configurations, formatting, clippy, release build and docs. Four added regression tests passed on both pristine and candidate code, and the evaluator checked 3,434,040 semantic cases in each of four ANSI/Unicode feature combinations.

The external autoresearch trajectory and exact source/metric bindings are available in [Weco Observe](https://dashboard.weco.ai/share/tH1Cu47EislOBE6MD4X2oVDCzlU0WCv1).

The measurements and Observe source snapshots correspond to research commit `13501c50490950db912aa6cdba672825939708e1`. Follow-up commit `b63f12304da138f3ba40c9a56de858a2a057c485` only moves the same four regression tests into `src/utils.rs`; the measured implementation is unchanged. The seven feature configurations, formatting, clippy, release build and docs were validated again for this revision. The benchmark was not motivated by an application profile.


tests/measure_text_width.rs
```diff
@@ -0,0 +1,33 @@
+#![cfg(all(feature = "std", feature = "ansi-parsing", feature = "unicode-width"))]
```
djc: Definitely don't want a new file for this.

src/utils.rs
```diff
@@ -115,6 +115,12 @@ pub fn set_true_colors_enabled_stderr(val: bool) {
 pub fn measure_text_width(s: &str) -> usize {
     #[cfg(feature = "ansi-parsing")]
     {
+        let printable_ascii = s
+            .bytes()
+            .fold(true, |ok, b| ok & (0x20..=0x7e).contains(&b));
```
djc, line 120: I think this can just be `all()` instead of `fold()`? Isn't there a `std` method on `u8` with a more readable variant of `& (0x20..=0x7e)`?

djc (commented):
Did this actually show up as some kind of bottleneck in your application?

djc (approved):

PR #300, opened 2026-09-23
ci: simplify MSRV job configuration

xtqqczze:
- Refactor the MSRV job to avoid an unnecessary matrix build
- Use `cargo-minimal-versions` to check the minimum supported dependencies and Rust version, including all feature combinations with `--feature-powerset` since the additional checks are cheap
- Check all Tier 1 targets in the MSRV job
- Remove redundant minimal version resolution job (it's incorporated into MSRV now)


djc:
Can you do one commit per bullet?

xtqqczze:
> Can you do one commit per bullet?

OK, I will make sure CI passes on each commit as well.

